PR #3, opened 2016-10-15
Removed optional arguments for consistency with final example code

crunchex:
The line description only refers to two values, the speeds of the left and right treads. A reader might be confused by the last two arguments (which are for left and right acceleration) as they disappear in the final example code.


MarkWesleyAtAnki:
Thanks for your contribution; we’re working on finalizing our contributor agreement - hopefully it should be fairly straightforward. We’ll let you know when that's ready yo go - in the meantime we'll leave this PR open.

MarkWesleyAtAnki:
drive_wheels is no longer used by this example.

PR #6, opened 2016-10-18
Prevent Android warning when iOS device is connected and running Cozmo

msintov:
Prevent Android warning when iOS device is connected and running Cozmo.

### DIFF
--- a/src/cozmo/run.py
+++ b/src/cozmo/run.py
@@ -294,11 +294,11 @@ class FirstAvailableConnector(DeviceConnector):
             return android_result
 
         ios = IOSConnector()
-        iOS_result = await self._do_connect(ios, *conn_args)
-        if not isinstance(iOS_result, BaseException):
-            return iOS_result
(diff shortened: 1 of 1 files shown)


src/cozmo/run.py
```diff
@@ -289,16 +289,17 @@ def __init__(self):
...
+        android_result = await self._do_connect(android, *conn_args)
+        if not isinstance(android_result, BaseException):
+            return android_result
 
         ios = IOSConnector()
-        result = await self._do_connect(ios, *conn_args)
-        if not isinstance(result, BaseException):
-            return result
-        logger.warn('No iOS device found running Cozmo: %s' % result)
+        iOS_result = await self._do_connect(ios, *conn_args)
```
MarkWesleyAtAnki: minor nitpick: we use lower snake case everywhere else for variable names so this feels slightly inconsistent

msintov: Fixed in 2fbd7d2

PR #7, opened 2016-10-18
Fix typo in abort_all_actions

MarkWesleyAtAnki:
Fix robotId vs robotID typo in Robot.abort_all_actions that made this function uncallable

### DIFF
--- a/src/cozmo/robot.py
+++ b/src/cozmo/robot.py
@@ -725,7 +725,7 @@ class Robot(event.Dispatcher):
         Abort / Cancel any action that is currently either running or queued within the engine
         '''
         # RobotActionType.UNKNOWN is a wildcard that matches all actions when cancelling.
-        msg = _clad_to_engine_iface.CancelAction(robotId=self.robot_id,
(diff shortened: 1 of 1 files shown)


src/cozmo/robot.py
```diff
@@ -725,7 +725,7 @@ def abort_all_actions(self):
         Abort / Cancel any action that is currently either running or queued within the engine
         '''
         # RobotActionType.UNKNOWN is a wildcard that matches all actions when cancelling.
-        msg = _clad_to_engine_iface.CancelAction(robotId=self.robot_id,
+        msg = _clad_to_engine_iface.CancelAction(robotID=self.robot_id,
```
msintov: Does the method above this one, set_robot_volume, have the same bug with robotId?

MarkWesleyAtAnki: No, unfortunately it's not completely consistent in the clad files - SetRobotVolume is declared as:
```message SetRobotVolume {
  uint_8   robotId,
  float_32 volume
}```

PR #10, opened 2016-10-21
Support skipping and clearing idle animation after face enrollment

MarkWesleyAtAnki:
Doing Face.name_face with the default sequence leaves an idle animation looping on Cozmo, so:
a) Allow that to be skipped (via passing in play_anim=False)
b) Allow idle animation to be set/cleared afterwards

Fixed a couple of existing issues with EnrollNamedFace docstrings being incorrectly generated

### DIFF
--- a/src/cozmo/faces.py
+++ b/src/cozmo/faces.py
@@ -33,7 +33,7 @@ observed by adding handlers there.
 __all__ = ['FACE_VISIBILITY_TIMEOUT',
            'EvtErasedEnrolledFace', 'EvtFaceAppeared', 'EvtFaceDisappeared',
            'EvtFaceIdChanged', 'EvtFaceObserved', 'EvtFaceRenamed',
-           'Face',
+           'EnrollNamedFace', 'Face',
            'erase_all_enrolled_faces', 'erase_enrolled_face_by_id',
            'update_enrolled_face_by_id']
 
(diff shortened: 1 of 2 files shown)


src/cozmo/faces.py
```diff
@@ -341,18 +348,20 @@ def _recv_msg_robot_erased_enrolled_face(self, evt, *, msg):
...
     #### Commands ####
 
-    def name_face(self, name):
+    def name_face(self, name, play_anim=True):
         '''Assign a name to this face. Cozmo will remember this name between SDK runs.
 
         Args:
             name (string): The name that will be assigned to this face
-
+            play_anim (bool): True plays a scanning animation,
```
msintov: Could reword: "If True then play a scanning animation. If False then ...'

src/cozmo/robot.py
```diff
@@ -931,6 +931,26 @@ def play_anim_trigger(self, trigger, loop_count=1):
...
         return action
 
+    def set_idle_animation(self, anim_trigger):
+        '''Set the Idle Animation on Cozmo
+
+        Args:
+            anim_trigger (:class:`cozmo.anim.Triggers`): The animation trigger to set
+                Note: :attr:`cozmo.anim.Triggers.Count` will clear all idle animations
+        Raises:
+            :class:`ValueError` if supplied an invalid idle_animation trigger.
```
msintov: Should 'idle_animation' be instead 'idle animation'?

src/cozmo/robot.py
```diff
@@ -931,6 +931,26 @@ def play_anim_trigger(self, trigger, loop_count=1):
         self._action_dispatcher._send_single_action(action)
         return action
 
+    def set_idle_animation(self, anim_trigger):
+        '''Set the Idle Animation on Cozmo
+
+        Args:
+            anim_trigger (:class:`cozmo.anim.Triggers`): The animation trigger to set
+                Note: :attr:`cozmo.anim.Triggers.Count` will clear all idle animations
```
msintov: Could end this Note with a period: "will clear all idle animations."

src/cozmo/faces.py
```diff
@@ -188,7 +195,7 @@ class Face(event.Dispatcher):
     which face it is looking at.
     '''
 
-    #: callable: The factory function to return a :class:`ErollNamedFace`
+    #: callable: The factory function to return an :class:`~cozmo.faces.EnrollNamedFace`
```
gwatts: you shouldn't need a fully qualified name if it's in the same scope. problem before was just the typo in `EnrollNamedFace`

src/cozmo/robot.py
```diff
@@ -931,6 +931,26 @@ def play_anim_trigger(self, trigger, loop_count=1):
         self._action_dispatcher._send_single_action(action)
         return action
 
+    def set_idle_animation(self, anim_trigger):
+        '''Set the Idle Animation on Cozmo
```
gwatts: i think some additional documentation on what "idle" means would be useful here... does it automatically restart the animation if it goes un-idle and then back again? etc.

src/cozmo/faces.py
gwatts: what's a "scanning animation"?  whatever it is, i think it needs a clearer description.. if there's not space here, then perhaps add it to the module and/or class docs

src/cozmo/robot.py
MarkWesleyAtAnki: added

PR #12, opened 2016-10-24
COZMO-6231-Added docstrings for the robot properties that were missing them

MarkWesleyAtAnki:
COZMO-6231 Added docstrings for the robot properties that were missing them

### DIFF
--- a/src/cozmo/objects.py
+++ b/src/cozmo/objects.py
@@ -141,7 +141,7 @@ class ObservableObject(event.Dispatcher):
 
         #: int: The robot's timestamp of the last observed event.
         #: ``None`` if the object has not yet been observed.
-        #: In milliseconds relative to robot epoch
+        #: In milliseconds relative to robot epoch.
         self.last_observed_robot_timestamp = None
 
         #: :class:`~cozmo.util.ImageBox`: The ImageBox defining where the
@@ -283,9 +283,9 @@ class LightCube(ObservableObject):
         #: float: The time the object was last tapped
         self.last_tapped_time = None
 
-        #: int: The robot's timestamp of the last tapped event
-        #: ``None`` if the cube wasn't tapped yet
-        #: In milliseconds relative to robot epoch
+        #: int: The robot's timestamp of the last tapped event.
+        #: ``None`` if the cube wasn't tapped yet.
+        #: In milliseconds relative to robot epoch.
         self.last_tapped_robot_timestamp = None
 
 
--- a/src/cozmo/objects.py
+++ b/src/cozmo/objects.py
@@ -141,6 +141,7 @@ class ObservableObject(event.Dispatcher):
 
         #: int: The robot's timestamp of the last observed event.
         #: ``None`` if the object has not yet been observed.
+        #: In milliseconds relative to robot epoch
         self.last_observed_robot_timestamp = None
 
         #: :class:`~cozmo.util.ImageBox`: The ImageBox defining where the
@@ -283,6 +284,8 @@ class LightCube(ObservableObject):
         self.last_tapped_time = None
 
         #: int: The robot's timestamp of the last tapped event
+        #: ``None`` if the cube wasn't tapped yet
+        #: In milliseconds relative to robot epoch
         self.last_tapped_robot_timestamp = None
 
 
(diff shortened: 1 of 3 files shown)


src/cozmo/robot.py
```diff
@@ -483,14 +484,23 @@ def __init__(self, conn, robot_id, is_primary, **kw):
 
         self._action_dispatcher = self._action_dispatcher_factory(self)
 
+        #: :class:`cozmo.util.Speed` Speed of the left wheel
```
gwatts: The type should have a colon after it .. i.e... ``...util.Speed`: Speed of...``

src/cozmo/robot.py
```diff
@@ -483,14 +484,23 @@ def __init__(self, conn, robot_id, is_primary, **kw):
...
         self.left_wheel_speed = None
+        #: :class:`cozmo.util.Speed` Speed of the right wheel
         self.right_wheel_speed = None
-        self.light_height = None
+        #: :class:`cozmo.util.Distance` height of the lift from the ground (in :const:`MIN_LIFT_HEIGHT_MM` to :const:`MAX_LIFT_HEIGHT_MM` range)
```
gwatts: you can wrap these comments by also beginning subsequent lines with `#:`

src/cozmo/robot.py
```diff
@@ -483,14 +484,23 @@ def __init__(self, conn, robot_id, is_primary, **kw):
...
+        self.lift_height = None
+        #: float: the current battery_voltage (not linear, but < 3.5 is low)
```
gwatts: can just be "battery voltage" without the underscore, no?

src/cozmo/robot.py
```diff
@@ -473,6 +473,7 @@ def __init__(self, conn, robot_id, is_primary, **kw):
 
         self._is_ready = False
         self._pose = None
+        #: bool: is this the primary robot (always True currently)
```
msintov: Could reword this to say: "Specifies that this is the primary robot"

src/cozmo/robot.py
```diff
@@ -483,14 +484,23 @@ def __init__(self, conn, robot_id, is_primary, **kw):
...
+        self.lift_height = None
+        #: float: the current battery_voltage (not linear, but < 3.5 is low)
         self.battery_voltage = None
+        #: int: the ID of the object currently being carried (-1 if none)
         self.carrying_object_id = -1
+        #: int: the ID of the object on top of the object currently being carried (-1 if none)
         self.carrying_object_on_top_id = -1
+        #: int: the ID of the object the head is tracking to (-1 if none)
         self.head_tracking_object_id  = -1
+        #: int: the ID of the object that robot is localized to (-1 if none)
```
msintov: Could reword this to "that the robot is..."

src/cozmo/robot.py
msintov: Each of these descriptions should start with a capital letter to match the existing descriptions.

src/cozmo/robot.py
```diff
@@ -483,14 +484,23 @@ def __init__(self, conn, robot_id, is_primary, **kw):
...
         self.localized_to_object_id = -1
+        #: int: the timestamp for the last image seen (>= 0)
```
gwatts: This timestamp is another millisecond timestamp relative to the boot time of the robot, correct?  along the lines of `last_observed_robot_timestamp` and `last_tapped_robot_timestamp` in `object.py`

We should probably clarify the docs for those types to stay how they're generated.. would be nice if this one was also called `last_image_robot_timestamp` as well to be consistent though i guess that's a breaking change at this point

src/cozmo/objects.py
```diff
@@ -283,6 +284,8 @@ def __init__(self, *a, **kw):
         self.last_tapped_time = None
 
         #: int: The robot's timestamp of the last tapped event
+        #: ``None`` if the cube wasn't tapped yet
+        #: In milliseconds relative to robot epoch
```
gwatts: Needs punctuation - Sphinx will run these lines together and it doesn't look good without it (plus i've been trying to be good about punctuation for Sphinx docs in general)

MarkWesleyAtAnki:
@gwatts I've added full-stops after after all those lines

PR #13, opened 2016-10-25
repo builds should not depend on explicit cozmoclad

gwatts:
Only production builds should be pinned to an explicit version of
comzmoclad; performing a build from a checkout should use whichever
version is installed (though subject to the usual runtime checks).

### DIFF
--- a/src/cozmo/version.py
+++ b/src/cozmo/version.py
@@ -17,5 +17,5 @@ __version__ = "0.7.1.dev0"
 # Specify the version of cozmoclad that this package requires
 # Releases of the Cozmo package must specify an exact cozmoclad release
 # to ensure compatibility with a specific release of the ios/android app.
-#__cozmoclad_version__ = None
-__cozmoclad_version__ = "1.0.1"
+__cozmoclad_version__ = None
+#__cozmoclad_version__ = "1.0.1"


src/cozmo/version.py
```diff
@@ -17,5 +17,5 @@
 # Specify the version of cozmoclad that this package requires
 # Releases of the Cozmo package must specify an exact cozmoclad release
 # to ensure compatibility with a specific release of the ios/android app.
-#__cozmoclad_version__ = None
-__cozmoclad_version__ = "1.0.1"
+__cozmoclad_version__ = None
```
MarkWesleyAtAnki: We should document this in README.md - I suspect that anyone forking this repo without following the cozmo installation steps, or even if they did follow those steps once, will run into issues - they'd need to know to manually install the correct cozmoclad module

gwatts: 99% of the time they shouldn't need to manually install as it will pull the latest published version of the cozmoclad package if they don't already have one installed, and that'll match the latest app in the app store

MarkWesleyAtAnki: That will break with a clad mismatch on the next app update, i'll add a task that we need to document how to use the repo.

PR #16, opened 2016-10-26
Added Charger object support to SDK

MarkWesleyAtAnki:
Added charger factory and object
Added docsstrings for factory objects in world.py
Send RequestAvailableObjects on robot initialize - and handle AvailableObjects message that is returned, this gives us info on connected cubes at the start instead of having to wait to visually see them.
Upgraded "Tap event received for unknown object ID" back to a warning as we should always know about that cube now
Fixed robot.py factory docstrings (they were missing a colon before class)
Made robot keep track of it's origin_id for poses
Added GoToObject support for driving to within X mm of a cube or charger
Added go_to_object_test.py example that shows usage of GoToObject action for finding a cube and driving up to it
Added drive_to_charger_test.py example that shows usage of accessing the charger and driving near to it.

### DIFF
--- a/examples/drive_to_charger_test.py
+++ b/examples/drive_to_charger_test.py
@@ -1,0 +1,108 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2016 Anki, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License in the file LICENSE.txt or at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+'''Make Cozmo drive towards his charger.
+
+The script shows an example of accessing the charger object from
+Cozmo's world, and driving towards it.
+'''
+
+import asyncio
+import sys
+import time
+
+import cozmo
+from cozmo.util import degrees, distance_mm, speed_mmps
+
+
+def find_charger(robot):
+    '''The core of the follow_faces program'''
+
+    # If the robot was on the charger, drive them forward and clear of the charger
+    if robot.is_on_charger:
+        # drive off the charger
+        robot.drive_off_charger_contacts().wait_for_completed()
+        robot.drive_straight(distance_mm(100), speed_mmps(50)).wait_for_completed()
+        # Start moving the lift down
+        robot.move_lift(-3)
+        # turn around to look at the charger
+        robot.turn_in_place(degrees(180)).wait_for_completed()
+        # Tilt the head to be level
+        robot.set_head_angle(degrees(0)).wait_for_completed()
+        # wait half a second to ensure Cozmo has seen the charger
+        time.sleep(0.5)
+        # drive backwards away from the charger
+        robot.drive_straight(distance_mm(-60), speed_mmps(50)).wait_for_completed()
+
+    # try to find the charger
+    charger = robot.world.charger
+
+    if charger:
+        if charger.pose.origin_id == robot.pose.origin_id:
+            print("Cozmo already knows where the charger is!")
+        else:
+            # Cozmo knows about the charger, but the pose is not based on the
+            # same origin as the robot (e.g. the robot was moved since seeing
+            # the charger) so try to look for the charger first
+            charger = None
+
+    if not charger:
+        # Tell Cozmo to look around for the charger
+        look_around = robot.start_behavior(cozmo.behavior.BehaviorTypes.LookAroundInPlace)
+        try:
+            charger = robot.world.wait_for_observed_charger(timeout=30)
+            print("Found charger: %s" % charger)
+        except asyncio.TimeoutError:
+            print("Didn't see the charger")
+        finally:
+            # whether we find it or not, we want to stop the behavior
+            look_around.stop()
+
+    if not charger:
+        charger = robot.world.charger
+        if charger:
+            if charger.pose.origin_id == robot.pose.origin_id:
+                print("Cozmo now knows where the charger is!")
+            else:
+                print("Attempting to drive to last known location of the charger")
+
+    if charger:
+        # Attempt to drive near to the charger, and then stop.
+        action = robot.go_to_object(charger, distance_mm(65.0))
+        action.wait_for_completed()
+        print("Completed action: result = %s" % action)
+        print("Done.")
+
+
+def run(sdk_conn):
+    '''The run method runs once the Cozmo SDK is connected.'''
+    robot = sdk_conn.wait_for_robot()
+
+    try:
+        find_charger(robot)
+
+    except KeyboardInterrupt:
+        print("")
+        print("Exit requested by user")
+
+
+if __name__ == '__main__':
+    cozmo.setup_basic_logging()
+    cozmo.robot.Robot.drive_off_charger_on_connect = False  # Cozmo can stay on charger for now
+    try:
+        cozmo.connect_with_tkviewer(run, force_on_top=True)
+    except cozmo.ConnectionError as e:
+        sys.exit("A connection error occurred: %s" % e)
--- a/examples/drive_to_charger_test.py
+++ b/examples/drive_to_charger_test.py
@@ -28,8 +28,8 @@ import cozmo
 from cozmo.util import degrees, distance_mm, speed_mmps
 
 
-def find_charger(robot):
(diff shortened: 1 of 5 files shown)


examples/drive_to_charger_test.py
```diff
@@ -0,0 +1,108 @@
...
+import asyncio
+import sys
+import time
+
+import cozmo
+from cozmo.util import degrees, distance_mm, speed_mmps
+
+
+def find_charger(robot):
+    '''The core of the follow_faces program'''
```
msintov: Do we want to keep this 'follow_faces' docstring?

src/cozmo/objects.py
```diff
@@ -95,6 +95,17 @@ class EvtObjectAppeared(event.Event):
     pose = 'The cozmo.util.Pose defining the position and rotation of the object.'
 
 
+class EvtObjectAvailable(event.Event):
+    '''Triggered when the engine reports that an object is available (i.e. exists)
```
msintov: Need a period at the end of this first line.

src/cozmo/objects.py
```diff
@@ -95,6 +95,17 @@ class EvtObjectAppeared(event.Event):
...
+
+    This will usually occur at the start of the program in response to the SDK
+    sending RequestAvailableObjects to the engine.
+    '''
+    obj = 'The object that is available'
+    updated = 'A set of field names that have changed'
+    pose = 'The cozmo.util.Pose defining the position and rotation of the object.'
```
msintov: The strings assigned to obj and updated don't end with periods but pose does. Should the period at the end of the pose string be removed?

src/cozmo/objects.py
```diff
@@ -365,6 +398,12 @@ def set_lights_off(self):
         self.set_lights(lights.off_light)
 
 
+class Charger(ObservableObject):
+    '''Cozmo's charger object, can be seen and driven towards'''
```
msintov: Maybe reword: "Cozmo's charger object, which the robot can see and drive toward."

src/cozmo/robot.py
```diff
@@ -1141,6 +1172,28 @@ def go_to_pose(self, pose, relative_to_robot=False):
         self._action_dispatcher._send_single_action(action)
         return action
 
+    def go_to_object(self, object, distance_from_object):
+        '''Tells Cozmo to drive to the specified object.
+
+        Args:
+            object (:class:`cozmo.objects.ObservableObject`): The destination object.
+            distance_from_object (:class:`cozmo.util.Distance`): The distance from
```
msintov: Maybe reword the distance_from_object description like this:
'The distance from the object to stop. This is the distance between the origins. For instance, the distance from the robot's origin (between Cozmo's two front wheels) to the cube's origin (at the center of the cube) is ~40mm.

src/cozmo/world.py
```diff
@@ -136,7 +154,17 @@ def _allocate_object_from_msg(self, msg):
...
             return cube
 
-        elif (msg.objectFamily == _clad_to_game_cozmo.ObjectFamily.CustomObject):
+        elif msg.objectFamily == _clad_to_game_cozmo.ObjectFamily.Charger:
+            charger = self.charger_factory(self.conn, self, msg.objectID, dispatch_parent=self)
+            if self.charger:
+                logger.error('Allocating multiple chargers: existing charger=%s msg=%s', self.charger, msg)
+            self.charger = charger
+            self._objects[charger.object_id] = charger
+            charger._robot = self.robot  # XXX this will move if/when we have multi-robot support
```
msintov: Should 'XXX' be a TODO instead?

examples/drive_to_charger_test.py
```diff
@@ -0,0 +1,108 @@
...
+            charger = robot.world.wait_for_observed_charger(timeout=30)
+            print("Found charger: %s" % charger)
+        except asyncio.TimeoutError:
+            print("Didn't see the charger")
+        finally:
+            # whether we find it or not, we want to stop the behavior
+            look_around.stop()
+
+    if not charger:
+        charger = robot.world.charger
```
gwatts: why would `charger` be `None` at this point, yet `robot.world.charger` be set?

examples/drive_to_charger_test.py
```diff
@@ -0,0 +1,108 @@
...
+            else:
+                print("Attempting to drive to last known location of the charger")
+
+    if charger:
+        # Attempt to drive near to the charger, and then stop.
+        action = robot.go_to_object(charger, distance_mm(65.0))
+        action.wait_for_completed()
+        print("Completed action: result = %s" % action)
+        print("Done.")
+
```
gwatts: all this `if charger` stuff feels a bit convoluted for an example

src/cozmo/objects.py
```diff
@@ -186,6 +197,28 @@ def _observed_timeout(self):
         self._is_visible = False
         self.dispatch_event(EvtObjectDisappeared, obj=self)
 
+    def _handle_available_object(self, available_object):
+        # triggered when engine sends available objects
+        # as a response to a RequestAvailableObjects message
+        if self.last_observed_robot_timestamp and \
+                (self.last_observed_robot_timestamp > available_object.lastObservedTimestamp):
```
gwatts: Better style to wrap the whole if condition in parentheses and drop the backslash at the end of the first line

src/cozmo/objects.py
```diff
@@ -186,6 +197,28 @@ def _observed_timeout(self):
...
+
+        self._pose = util.Pose(available_object.pose.x, available_object.pose.y, available_object.pose.z,
+                               q0=available_object.pose.q0, q1=available_object.pose.q1,
+                               q2=available_object.pose.q2, q3=available_object.pose.q3,
+                               origin_id=available_object.pose.originID)
+
+        self.dispatch_event(EvtObjectAvailable,
+                            obj=self,
+                            updated=changed_fields,
+                            pose = self._pose)
```
gwatts: stray whitespace around the `=`

src/cozmo/objects.py
gwatts: should we stick with "observe" rather than "see" for the sake of consistency?

src/cozmo/robot.py
```diff
@@ -1141,6 +1172,28 @@ def go_to_pose(self, pose, relative_to_robot=False):
...
+                the object to stop at - this is the distance between the origins so
+                e.g. from between Cozmo's 2 front wheels to the center of a cube
+                is ~40mm
+        Returns:
+            A :class:`cozmo.robot.GoToObject` action object which can be queried
+                to see when it is complete.
+        '''
+        if not isinstance(object, objects.ObservableObject):
+            raise TypeError("Invalid charger supplied")
```
gwatts: Presumably don't mean charger here, since it can be any  observable object..

Perhaps "Target must be an observable object" instead?

src/cozmo/world.py
MarkWesleyAtAnki: I've always used TODO in the past, but seems that XXX is commonly used to point out odd but not necessarily broken (or to-be-done) things.

examples/drive_to_charger_test.py
MarkWesleyAtAnki: I will simplify, unfortunately at the cost of it working less reliably.

examples/drive_to_charger_test.py
MarkWesleyAtAnki: fixed

src/cozmo/objects.py
MarkWesleyAtAnki: fixed

src/cozmo/objects.py
MarkWesleyAtAnki: Was following the pattern from the other events, but yeah i guess they should be consistent. Fixed.

src/cozmo/objects.py
MarkWesleyAtAnki: Fixed

src/cozmo/objects.py
MarkWesleyAtAnki: Fixed

src/cozmo/robot.py
MarkWesleyAtAnki: Fixed

src/cozmo/world.py
gwatts: Yeah `TODO` implies that there's actionable work to be done IMO; if you're bored and grep the source tree for it, you should find things  you can work on.

`XXX` means warning, might change, heads up, etc

src/cozmo/world.py
```diff
@@ -102,6 +116,10 @@ def __init__(self, conn, robot, **kw):
 
         self.light_cubes = {}
 
+        #: :class:`cozmo.objects.Charger`: Cozmo's charger.
```
gwatts: there is no `cozmo.objects.Charger` class... presumably mean `ObservableObject `

MarkWesleyAtAnki: There is - I added cozmo.objects.Charger as part of this PR

PR #18, opened 2016-10-27
Add upgrade instruction to docs

msintov:


msintov:
Added "SDK Upgrade" subheading in af961fe

PR #19, opened 2016-10-28
Auto docking feature

scas-mdd:
Based on the example added by @MarkWesleyAtAnki in 0642337 
See also thread https://forums.anki.com/t/cozmo-self-parking-on-charger-when-battery-is-low/414/50

I hope others can also use this as an inspiration point and I will try to make a better and precise version of this in the following days.


MarkWesleyAtAnki:
Thanks for your contribution, it's not something we want to merge in at this stage as we already have the similar drive_to_charger example and we'd prefer to extend that in the future with docking if-and-when we can make it sufficiently clean and reliable, rather than have to support additional example programs for each future SDK release. In this case I think your auto_dock example would be best maintained by you on your fork, and you can e.g. post it in the Showcase section of our forums.

PR #21, opened 2016-10-28
Same links available in getstarted.rst but added here for thoroughness

jfhardy:
Links for free source code editors for Mac/Windows/Linux

### DIFF
--- a/docs/source/tutorial-beginner.rst
+++ b/docs/source/tutorial-beginner.rst
@@ -104,7 +104,7 @@ Example 1 - Drive Straight
 
 For your first program, you will tell Cozmo to drive in a straight line for three seconds. This program will give you a simple overview of the programming process, as well as some of the building blocks necessary for the programs to work.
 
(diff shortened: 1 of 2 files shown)


docs/source/tutorial-beginner.rst
```diff
@@ -104,7 +104,7 @@ Example 1 - Drive Straight
 
 For your first program, you will tell Cozmo to drive in a straight line for three seconds. This program will give you a simple overview of the programming process, as well as some of the building blocks necessary for the programs to work.
 
-1. In your source code editor, create a new document (*File -> New Document*).
+1. In your source code editor, create a new document (*File -> New Document*). Free source code editors, such as `Atom <https://atom.io>`_ , `Sublime <https://www.sublimetext.com>`_ , or `TextWrangler <http://www.barebones.com/products/textwrangler>`_ can be found online. Anki does not provide tech support for third-party source code editors.
```
MarkWesleyAtAnki: I think we should suggest https://www.jetbrains.com/pycharm/ (The free community edition of it) as the 1st option as it's one of the most common python editors.

jfhardy: OK, added in both areas where source code editors are suggested.

PR #24, opened 2016-10-31
If This Then That examples

msintov:
[DO NOT MERGE TO MASTER UNTIL WE ARE READY AND NEW IMAGES ADDED.]

This PR contains three If This Then That examples (ESPN, Gmail and Stocks) as well as helper file if_this_then_that_helpers.py which contains shared code.

The ESPN example reacts to in-game updates for your team; the Gmail example reacts to new Gmails being received; and the Stocks example reacts to the the stock price increasing for the ticker symbol you specify. All require recipes to be set up on IFTTT and the steps to do so are documented.

Awaiting images for Gmail and Stocks example.


msintov:
Pull request updated to use aiohttp for two of the IFTTT examples (gmail and stocks) and flask with a queue for one example (sports).

In a separate pull request my plan is to move perform_operation_off_charger and backup_onto_charger methods (both sync and async) so they can be shared between these examples and alarm_clock.py.

This PR is ready for review.

examples/if_this_then_that/common/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+
+import sys
+
+import cozmo
+
+
+try:
+    from PIL import Image, ImageDraw, ImageFont
+except ImportError:
+    sys.exit("Cannot import from PIL: Do `pip3 install Pillow` to install")
```
MarkWesleyAtAnki: Can you change this to the following (I think i updated all the other examples to add "--user" after you started working on this):
sys.exit("Cannot import from PIL. Do `pip3 install --user Pillow` to install")

examples/if_this_then_that/common/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def display_image_file_on_face(self, image_name):
+        # load image and convert it for display on cozmo's face
+        image = Image.open(image_name)
+
+        # resize to fit on Cozmo's face screen
+        resized_image = image.resize(cozmo.oled_face.dimensions(), Image.NEAREST)
+
+        # convert the image to the format used by the oled screen
+        face_image = cozmo.oled_face.convert_image_to_screen_data(resized_image,
```
MarkWesleyAtAnki: Not a huge issue for now, so probably not worth changing at this stage, but you're combining the loading with the display - so it will load and convert each time - might have been cleaner if the helper just returned face_image, you cache those at the program start, and then make the relevant display_oled_face_image calls later?

examples/if_this_then_that/ifttt_gmail.py
```diff
@@ -0,0 +1,159 @@
...
+    '''
+
+    json_object = await request.json()
+
+    # Extract the name of the email sender.
+    from_email_address = json_object["FromAddress"]
+
+    # Use a regular expression to break apart pieces of the email address
+    match_object = re.search(r'([\w.]+)@([\w.]+)', from_email_address)
+    email_local_part = match_object.group(1)
```
MarkWesleyAtAnki: Minor/NitPick/Don'tFixNow. I assume this just gets the bit before the @ symbol - so e.g. name.name0123456789 - is it possible to get the associated name (like when you get the email there is often a corresponding name like `<"john doe" john.doe0123456789@....>`

examples/if_this_then_that/ifttt_sports_flask.py
```diff
@@ -0,0 +1,198 @@
...
+        b) On ifttt.com, on your applet page, click “Check now”. See that IFTTT confirms that the applet
+            was checked.
+        c) Wait for new in-game updates for your team and see Cozmo react! Cozmo should roll off the charger, raise
+            and lower his lift, show an image on his face and speak the in-game update.
+'''
+
+import json
+import queue
+import sys
+sys.path.append('../')
```
MarkWesleyAtAnki: Minor/For-later: i assume that's specifically for importing flask_helpers, might be better next to it. We should look at reworking examples folders in general to make it easier to understand and the order to try them in etc.

examples/if_this_then_that/common/__init__.py
msintov: Fixed in e29b042

examples/if_this_then_that/common/__init__.py
msintov: Good recommendation.

examples/if_this_then_that/ifttt_gmail.py
msintov: It is not possible to get the associated name from IFTTT, unfortunately.

examples/if_this_then_that/ifttt_sports_flask.py
msintov: Moved the append line next to flask_helpers import in 8d6e5a1

PR #25, opened 2016-11-03
Log version numbers at startup

gwatts:
Replaced the ambiguous "UI connected" message with something more
meaningful and record the SDK version along with the version of
cozmoclad and the connected build version.

Lower the animations loaded count to a debug as it's generally not a
useful number.

### DIFF
--- a/src/cozmo/conn.py
+++ b/src/cozmo/conn.py
@@ -356,7 +356,7 @@ class CozmoConnection(event.Dispatcher, clad_protocol.CLADProtocol):
         self._is_ui_connected = True
         self.dispatch_event(EvtConnected, conn=self)
         self.anim_names.refresh()
-        logger.info('App connection established.  sdk_version=%s  '
+        logger.info('App connection established. sdk_version=%s '
                 'cozmoclad_version=%s app_build_version=%s',
                 version.__version__, cozmoclad.__version__, msg.buildVersion)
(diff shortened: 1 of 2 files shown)


src/cozmo/conn.py
```diff
@@ -356,7 +356,9 @@ def _handle_ui_device_connected(self, msg):
         self._is_ui_connected = True
         self.dispatch_event(EvtConnected, conn=self)
         self.anim_names.refresh()
-        logger.info("UI device connected")
+        logger.info('App connection established.  sdk_version=%s  '
+                'cozmoclad_version=%s app_build_version=%s',
```
MarkWesleyAtAnki: spacing/formatting seems a bit random, and will wrap with a terminal narrower than 160 chars
`2016-11-02 17:43:40,510 cozmo.general INFO     App connection established.  sdk_version=0.8.1.dev0  cozmoclad_version=1.0.2 app_build_version=00001.00000.00002`

gwatts: i can remove the extra space from the end of the first line, but the 2 spaces afte rthe period was deliberate.

Not too worried about line wrapping; it's a log file, not an interactive prompt.. i'd rather have related data grouped together on a single line than split it across multiple disconnected lines as analyzing it later becomes much more difficult.

PR #27, opened 2016-11-07
Allow SDK connection if CLAD and major.minor version matches

MarkWesleyAtAnki:
On CLAD match but version mismatch, check if the major.minor still match (i.e. this is just a patch version change) and if so allow it, but warn user that they should upgrade app or SDK as appropriate.
Added link to the downloads and suggestion that they should download the latest examples when upgrading the SDK.

### DIFF
--- a/src/cozmo/conn.py
+++ b/src/cozmo/conn.py
@@ -308,13 +308,13 @@ class CozmoConnection(event.Dispatcher, clad_protocol.CLADProtocol):
                 if cozmoclad.__build_version__ < msg.buildVersion:
                     # App is newer
                     logger.warning(
-                        'Your App is newer than your SDK: '
-                        'We recommend that you upgrade by calling command: '
-                        '"pip3 install --user --ignore-installed cozmo" '
+                        'Your app is newer than your SDK: '
+                        'We recommend that you upgrade your SDK by calling: '
+                        '"pip3 install --user --upgrade cozmo" '
                         'and downloading the latest examples from http://cozmosdk.anki.com/docs/downloads.html')
                 else:
                     logger.warning(
-                        'Your App is older than your SDK: '
+                        'Your app is older than your SDK: '
                         'We recommend that you download the latest app from the app store.')
 
         if clad_hashes_match and build_versions_match:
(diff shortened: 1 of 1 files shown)


src/cozmo/conn.py
```diff
@@ -299,6 +299,24 @@ def _handle_ui_device_connected(self, msg):
...
+                    # App is newer
+                    logger.warning(
+                        'Your App is newer than your SDK: '
+                        'We recommend that you upgrade by calling command: '
+                        '"pip3 install --user --ignore-installed cozmo" '
+                        'and downloading the latest examples from http://cozmosdk.anki.com/docs/downloads.html')
+                else:
+                    logger.warning(
+                        'Your App is older than your SDK: '
+                        'We recommend that you download the latest app from the app store.')
```
msintov: In "the latest app", app should be App to match the other references to App in this new code.

src/cozmo/conn.py
```diff
@@ -299,6 +299,24 @@ def _handle_ui_device_connected(self, msg):
...
+            build_major_version = msg.buildVersion.split(".")[0:2]
+            build_versions_match = (sdk_major_version == build_major_version)
+            if build_versions_match:
+                if cozmoclad.__build_version__ < msg.buildVersion:
+                    # App is newer
+                    logger.warning(
+                        'Your App is newer than your SDK: '
+                        'We recommend that you upgrade by calling command: '
+                        '"pip3 install --user --ignore-installed cozmo" '
+                        'and downloading the latest examples from http://cozmosdk.anki.com/docs/downloads.html')
```
msintov: We could make what is being upgraded more clear by saying: "We recommend that you upgrade your SDK by calling...."

Why do we recommend --ignore-installed here? in our docs we don't use that, and that does a reinstall, not upgrade.

"downloading" should be "download"

src/cozmo/conn.py
```diff
@@ -332,7 +350,8 @@ def _handle_ui_device_connected(self, msg):
                     # App is newer
                     logger.error(
                         'Please update your SDK to the newest version by calling command: '
-                        '"pip3 install --user --ignore-installed cozmo"')
+                        '"pip3 install --user --ignore-installed cozmo" '
```
msintov: Like above, why are we using --ignore-installed, since that does a reinstall? And should say "download" instead of "downloading".

src/cozmo/conn.py
gwatts: should probably be `--upgrade` instead

PR #28, opened 2016-11-08
COZMO-6586 Include already visible objects/faces

MarkWesleyAtAnki:
The "Wait for observed" functions (listed below) now all take in "include_already_visible" (defaulted to true) which includes already visible objects (as these do not create observation events, they would otherwise be ignored), you can obtain the old behavior by passing in include_already_visible=False:
Actions expose more info on success / failure
cube_stack example now more reliable, and includes information on action failure

wait_until_observe_num_objects
wait_for_observed_charger
wait_for_observed_face
wait_for_observed_light_cube

### DIFF
--- a/examples/cube_stack.py
+++ b/examples/cube_stack.py
@@ -41,12 +41,14 @@ def run(sdk_conn):
         current_action = robot.pickup_object(cubes[0])
         current_action.wait_for_completed()
         if current_action.failed:
-            print("Pickup Cube failed: code=%s reason=%s" % current_action.failure_reason)
+            code, reason = current_action.failure_reason
+            print("Pickup Cube failed: code=%s reason=%s" % (code, reason))
 
         current_action = robot.place_on_object(cubes[1])
         current_action.wait_for_completed()
         if current_action.failed:
-            print("Place On Cube failed: code=%s reason=%s" % current_action.failure_reason)
+            code, reason = current_action.failure_reason
+            print("Place On Cube failed: code=%s reason=%s" % (code, reason))
 
 if __name__ == '__main__':
     cozmo.setup_basic_logging()
--- a/examples/cube_stack.py
+++ b/examples/cube_stack.py
@@ -38,8 +38,15 @@ def run(sdk_conn):
     if len(cubes) < 2:
         print("Error: need 2 Cubes but only found", len(cubes), "Cube(s)")
     else:
-        robot.pickup_object(cubes[0]).wait_for_completed()
-        robot.place_on_object(cubes[1]).wait_for_completed()
+        current_action = robot.pickup_object(cubes[0])
+        current_action.wait_for_completed()
+        if current_action.failed:
+            print("Pickup Cube failed: code=%s reason=%s" % current_action.failure_reason)
+
+        current_action = robot.place_on_object(cubes[1])
+        current_action.wait_for_completed()
+        if current_action.failed:
+            print("Place On Cube failed: code=%s reason=%s" % current_action.failure_reason)
 
 if __name__ == '__main__':
     cozmo.setup_basic_logging()
--- a/src/cozmo/action.py
+++ b/src/cozmo/action.py
@@ -148,12 +148,12 @@ class Action(event.Dispatcher):
         return self._state in (ACTION_SUCCEEDED, ACTION_FAILED)
 
     @property
-    def succeeded(self):
+    def has_succeeded(self):
         '''bool: True if the action has succeeded.'''
         return self._state == ACTION_SUCCEEDED
 
     @property
-    def failed(self):
+    def has_failed(self):
         '''bool: True if the action has failed.'''
         return self._state == ACTION_FAILED
 
--- a/src/cozmo/action.py
+++ b/src/cozmo/action.py
@@ -100,6 +100,8 @@ class Action(event.Dispatcher):
         extra = self._repr_values()
         if len(extra) > 0:
             extra = ' '+extra
+        if self._state == ACTION_FAILED:
+            extra += " failure_reason=%s failure_code=%s" % (self._failure_reason, self._failure_code)
         return '<%s state=%s%s>' % (self.__class__.__name__, self.state, extra)
 
     def _repr_values(self):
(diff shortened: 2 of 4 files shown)


src/cozmo/action.py
```diff
@@ -146,6 +148,16 @@ def is_completed(self):
         return self._state in (ACTION_SUCCEEDED, ACTION_FAILED)
 
     @property
+    def succeeded(self):
```
gwatts: should be `is_succeeded` or perhaps `has_succeeded` to match the pattern of the other boolean properties

src/cozmo/action.py
```diff
@@ -146,6 +148,16 @@ def is_completed(self):
...
+        '''bool: True if the action has succeeded.'''
+        return self._state == ACTION_SUCCEEDED
+
+    @property
+    def failed(self):
```
gwatts: `is/has_failed`

src/cozmo/objects.py
```diff
@@ -170,8 +170,8 @@ def __repr__(self):
         if len(extra) > 0:
             extra = ' '+extra
 
-        return '<%s is_visible=%s pose=%s%s>' % (self.__class__.__name__,
-                self.is_visible, self.pose, extra)
+        return '<%s id=%s is_visible=%s pose=%s%s>' % (self.__class__.__name__,
```
gwatts: let's stick to `object_id` here as `id` usually refers to Python object identifiers

examples/cube_stack.py
```diff
@@ -38,8 +38,15 @@ def run(sdk_conn):
...
-        robot.place_on_object(cubes[1]).wait_for_completed()
+        current_action = robot.pickup_object(cubes[0])
+        current_action.wait_for_completed()
+        if current_action.failed:
+            print("Pickup Cube failed: code=%s reason=%s" % current_action.failure_reason)
+
+        current_action = robot.place_on_object(cubes[1])
+        current_action.wait_for_completed()
+        if current_action.failed:
+            print("Place On Cube failed: code=%s reason=%s" % current_action.failure_reason)
```
gwatts: Might be better to log rather than print.. also clearer to break out code and reason:
```python
code, reason = current_action.failure_reason
cozmo.logger.error("Play on cube failed: code=%s reason=%s", code, reason)
```

src/cozmo/world.py
```diff
@@ -344,48 +344,81 @@ def recv_evt_face_disappeared(self, evt, *, face, **kw):
...
+            if (object_type is None) or isinstance(visible_object, object_type):
+                return visible_object
+        return None
+
+    def _find_visible_face(self):
+        for visible_face in self.visible_faces:
+            return visible_face
+        return None
+
+    async def wait_for_observed_light_cube(self, timeout=None, include_already_visible=True):
```
gwatts: can we call this `include_existing` so that it'd be the same if we applied it to other similar `wait_for...` style operators?

src/cozmo/world.py
```diff
@@ -344,48 +344,81 @@ def recv_evt_face_disappeared(self, evt, *, face, **kw):
...
         Args:
             timeout (float): Number of seconds to wait for a face to be
                 observed, or None for indefinite
+            include_already_visible (bool): whether or not to include faces
+                that are already visible.
         Returns:
             The :class:`cozmo.faces.Face` object that was observed.
         '''
+        if include_already_visible:
+            face = self._find_visible_face()
```
gwatts: can replace `_find_visible_face` with just `face = next(self.visible_faces, None)`

src/cozmo/world.py
```diff
@@ -344,48 +344,81 @@ def recv_evt_face_disappeared(self, evt, *, face, **kw):
...
         Args:
             timeout (float): Number of seconds to wait for a cube to be
                 observed, or None for indefinite
+            include_already_visible (bool): whether or not to include light cubes
+                that are already visible.
         Returns:
             The :class:`cozmo.objects.LightCube` object that was observed.
         '''
+        if include_already_visible:
+            obj = self._find_visible_object(objects.LightCube)
```
gwatts: if you want to eliminate another function..
```python
obj = next(filter(lambda obj: isinstance(obj, objects.LightCube), self.visible_objects), None)
```

src/cozmo/world.py
```diff
@@ -344,48 +344,81 @@ def recv_evt_face_disappeared(self, evt, *, face, **kw):
...
+            return visible_face
+        return None
+
+    async def wait_for_observed_light_cube(self, timeout=None, include_already_visible=True):
         '''Waits for one of the light cubes to be observed by the robot.
 
         Args:
             timeout (float): Number of seconds to wait for a cube to be
                 observed, or None for indefinite
+            include_already_visible (bool): whether or not to include light cubes
```
msintov: I'd change this text to: "Specifies whether to include..."

src/cozmo/world.py
```diff
@@ -344,48 +344,81 @@ def recv_evt_face_disappeared(self, evt, *, face, **kw):
...
         return evt.obj
 
-    async def wait_for_observed_face(self, timeout=None):
+    async def wait_for_observed_face(self, timeout=None, include_already_visible=True):
         '''Waits for a face to be observed by the robot.
 
         Args:
             timeout (float): Number of seconds to wait for a face to be
                 observed, or None for indefinite
+            include_already_visible (bool): whether or not to include faces
```
msintov: Again, could change to: "Specifies whether to include..."

src/cozmo/world.py
```diff
@@ -344,48 +344,81 @@ def recv_evt_face_disappeared(self, evt, *, face, **kw):
...
         return evt.face
 
-    async def wait_for_observed_charger(self, timeout=None):
+    async def wait_for_observed_charger(self, timeout=None, include_already_visible=True):
         '''Waits for a charger to be observed by the robot.
 
         Args:
             timeout (float): Number of seconds to wait for a charger to be
                 observed, or None for indefinite
+            include_already_visible (bool): whether or not to include chargers
```
msintov: Could change to: "Specifies whether to include..."

src/cozmo/world.py
```diff
@@ -402,6 +435,8 @@ def recv_evt_face_disappeared(self, evt, *, face, **kw):
                 this will cause only the selected object types to be counted.
             timeout (float): Maximum amount of time in seconds to wait for the
                 requested number of objects to be observed.
+            include_already_visible (bool): whether or not to include objects
```
msintov: Again, "Specifies whether to include..."

examples/cube_stack.py
MarkWesleyAtAnki: For the examples we've generally used print, vs the logger for the actual sdk

src/cozmo/world.py
MarkWesleyAtAnki: I feel that overcomplicates the code and makes it harder to read

PR #29, opened 2016-11-08
Add perform_off_charger context manager to Robot

gwatts:


src/cozmo/robot.py
```diff
@@ -392,6 +392,25 @@ def _encode(self):
...
+
+class PerformOffChargerContext(event.Dispatcher):
+    '''A helper class to provide a context manager to do operations while Cozmo is off charger.'''
+    def __init__(self, robot, **kw):
+        super().__init__(**kw)
+        self.robot = robot
+
+    async def __aenter__(self):
+        self.was_on_charger = self.robot.is_on_charger
+        self.was_on_charger = True
```
MarkWesleyAtAnki: Is this deliberate - the previous line's result is ignored? This would incorrectly always attempt to backup onto a charger

src/cozmo/robot.py
```diff
@@ -392,6 +392,25 @@ def _encode(self):
...
+        await self.robot.drive_off_charger_contacts().wait_for_completed()
```
MarkWesleyAtAnki: We could skip this action if not already on charger (it completes within about 2 frames otherwise so doesn't make a huge difference).

src/cozmo/robot.py
gwatts: ugh no.. it's because is_on_charger is returning False for my setup for some reason, and not for Michelle's.. will remove.

src/cozmo/robot.py
gwatts: the interesting thing is, this function is driving my robot off the charger even though is_on_charger is returning False

MarkWesleyAtAnki: That suggests to me that is_on_charger isn't being set properly - like a RobotStatus message hasn't made it back to SDK yet?

PR #32, opened 2016-11-08
Use robot.perform_off_charger now that we've moved that into robot

MarkWesleyAtAnki:
Print out how to set the alarm time (if none was provided)
Print out how to press CTRL-C to quit
Upped default backup_onto_charger timeout to 3 from 2 seconds (alarm clock was using 3, as there were occasional cases that it'd take too close to 2 seconds for comfort)

### DIFF
--- a/examples/alarm_clock.py
+++ b/examples/alarm_clock.py
@@ -234,7 +234,7 @@ def alarm_clock(robot):
     if alarm_time:
         print("Alarm set for %s" % alarm_time)
     else:
-        print("No Alarm time provided - usage e.g: 'alarm.clock.py 17:23' to set alarm for 5:23pm (input uses the 24 hour clock)")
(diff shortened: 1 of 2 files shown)


examples/alarm_clock.py
```diff
@@ -268,7 +234,8 @@ def alarm_clock(robot):
     if alarm_time:
         print("Alarm set for %s" % alarm_time)
     else:
-        print("No Alarm time provided")
+        print("No Alarm time provided - usage e.g: 'alarm.clock.py 17:23' to set alarm for 5:23pm (input uses the 24 hour clock)")
```
msintov: We could change this text as follows or similar:
"No Alarm time provided. Usage example: 'alarm.clock.py 17:23' to set alarm for 5:23 PM. (Input uses the 24-hour clock.)"

PR #34, opened 2016-11-08
Add synchronous example to the perform_off_charger docstring

msintov:


src/cozmo/robot.py
```diff
@@ -1273,7 +1273,13 @@ def perform_off_charger(self):
...
+        Synchronous example::
+
+            with robot.perform_off_charger():
+                action = robot.say_text("Hello")
+                action.wait_for_completed()
+
+        Asynchronous example::
 
             async with robot.perform_off_charger():
                 action = robot.say_text("Hello")
```
gwatts: seems redundant to duplicate the whole example .. could just say that synchronous code should use "with" instead of "async with"

msintov: The differences are both the "async with" and the "await". For copying and pasting code, I think the full example could be handy.

PR #36, opened 2016-11-23
Update the use of the observed messages as the syntax is now differen…

MarkWesleyAtAnki:
…t (rectangle now inside an img_rect member)


scopp:
👍  tested and it's working properly now.

PR #37, opened 2016-11-29
COZMO-4894 Expose Robot IMU data

MarkWesleyAtAnki:
Pull gyro and accelerometer IMU data out of RobotState message and store
Added Vector3 class for storing the above (basically renamed util.Position class, and made Position just a subclass of that)
Slightly more info in robot.pose property docstring
Added (optional) RobotState display annotation to remote_control example for verifying the IMU data
Fixed a couple of small bugs in remote control - updated default animations (2 were removed from 1.1.0), and removed the wait_for_completed on the drive_off_charger_contacts action (it isn't required, and would freeze the camera feed for a second)
Reordered display of remote control key mappings from 0..9 to 1..9,0 to match the layout of the keys on a keyboard.

### DIFF
--- a/examples/remote_control_cozmo.py
+++ b/examples/remote_control_cozmo.py
@@ -37,6 +37,11 @@ except ImportError:
     sys.exit("Cannot import from PIL: Do `pip3 install --user Pillow` to install")
 
 
+DEBUG_ANNOTATIONS_DISABLED = 0
+DEBUG_ANNOTATIONS_ENABLED_VISION = 1
+DEBUG_ANNOTATIONS_ENABLED_ALL = 2
+
+
 # Annotator for displaying RobotState (position, etc.) on top of the camera feed
 class RobotStateDisplay(cozmo.annotate.Annotator):
     def apply(self, image, scale):
@@ -45,7 +50,6 @@ class RobotStateDisplay(cozmo.annotate.Annotator):
         bounds = [3, 0, image.width, image.height]
 
         def print_line(text_line):
-            nonlocal bounds
             text = cozmo.annotate.ImageText(text_line, position=cozmo.annotate.TOP_LEFT, color='lightblue')
             text.render(d, bounds)
             TEXT_HEIGHT = 11
@@ -63,7 +67,7 @@ class RobotStateDisplay(cozmo.annotate.Annotator):
 
         # Display the Accelerometer and Gyro data for the robot
 
-        print_line('Acclrm: <%.1f, %.1f, %.1f>' % robot.accelerometer.x_y_z)
+        print_line('Accel: <%.1f, %.1f, %.1f>' % robot.accelerometer.x_y_z)
         print_line('Gyro: <%.1f, %.1f, %.1f>' % robot.gyro.x_y_z)
 
 
@@ -89,9 +93,6 @@ remote_control_cozmo = None
 _default_camera_image = create_default_image(320, 240)
 _is_mouse_look_enabled_by_default = False
 
-DEBUG_ANNOTATIONS_DISABLED = 0
-DEBUG_ANNOTATIONS_ENABLED_VISION = 1
-DEBUG_ANNOTATIONS_ENABLED_ALL = 2
 _display_debug_annotations = DEBUG_ANNOTATIONS_ENABLED_ALL
 
 
--- a/examples/remote_control_cozmo.py
+++ b/examples/remote_control_cozmo.py
@@ -32,9 +32,39 @@ except ImportError:
     sys.exit("Cannot import from flask: Do `pip3 install --user flask` to install")
 
 try:
-    from PIL import Image
+    from PIL import Image, ImageDraw
 except ImportError:
     sys.exit("Cannot import from PIL: Do `pip3 install --user Pillow` to install")
+
+
+# Annotator for displaying RobotState (position, etc.) on top of the camera feed
+class RobotStateDisplay(cozmo.annotate.Annotator):
+    def apply(self, image, scale):
+        d = ImageDraw.Draw(image)
+
+        bounds = [3, 0, image.width, image.height]
+
+        def print_line(text_line):
+            nonlocal bounds
+            text = cozmo.annotate.ImageText(text_line, position=cozmo.annotate.TOP_LEFT, color='lightblue')
+            text.render(d, bounds)
+            TEXT_HEIGHT = 11
+            bounds[1] += TEXT_HEIGHT
+
+        robot = self.world.robot
+
+        # Display the Pose info for the robot
+
+        pose = robot.pose
+        print_line('Pose: Pos = <%.1f, %.1f, %.1f>' % pose.position.x_y_z)
+        print_line('Pose: Rot quat = <%.1f, %.1f, %.1f, %.1f>' % pose.rotation.q0_q1_q2_q3)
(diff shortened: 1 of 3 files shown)


src/cozmo/robot.py
```diff
@@ -625,7 +638,9 @@ def anim_names(self):
 
     @property
     def pose(self):
-        """:class:`cozmo.util.Pose`: The current pose of cozmo relative to where he started when the engine was initialized.
+        """:class:`cozmo.util.Pose`: The current pose of cozmo relative to where it started
+        (when the engine was initialized, or any delocalizations since - e.g. whenever Cozmo is picked up)
```
gwatts: need a blank line between the summary line and the description

src/cozmo/robot.py
```diff
@@ -750,7 +765,9 @@ def _recv_msg_robot_state(self, evt, *, msg):
         self.left_wheel_speed  = util.speed_mmps(msg.leftWheelSpeed_mmps)
         self.right_wheel_speed = util.speed_mmps(msg.rightWheelSpeed_mmps)
         self.lift_height       = util.distance_mm(msg.liftHeight_mm)
-        self.battery_voltage           = msg.batteryVoltage
+        self.battery_voltage = msg.batteryVoltage
+        self.accelerometer = util.Vector3(msg.accel.x, msg.accel.y, msg.accel.z)
+        self.gyro = util.Vector3(msg.gyro.x, msg.gyro.y, msg.gyro.z)
```
gwatts: we should either align all these `=` symbols or none of them (either is fine with me)

examples/remote_control_cozmo.py
```diff
@@ -32,11 +32,41 @@
...
 
+# Annotator for displaying RobotState (position, etc.) on top of the camera feed
+class RobotStateDisplay(cozmo.annotate.Annotator):
+    def apply(self, image, scale):
+        d = ImageDraw.Draw(image)
+
+        bounds = [3, 0, image.width, image.height]
+
+        def print_line(text_line):
+            nonlocal bounds
```
gwatts: you don't technically need this `nonlocal` statement here as you're only referencing an existing variable, not assigning to it.. ie.. you'd need if it you'd done `bound=.. new bounds`, but as it is you're only mutating a value within an existing array, which as it doesn't exist locally will find by the usual scope rules

examples/remote_control_cozmo.py
```diff
@@ -32,11 +32,41 @@
...
 
 try:
-    from PIL import Image
+    from PIL import Image, ImageDraw
 except ImportError:
     sys.exit("Cannot import from PIL: Do `pip3 install --user Pillow` to install")
 
 
+# Annotator for displaying RobotState (position, etc.) on top of the camera feed
+class RobotStateDisplay(cozmo.annotate.Annotator):
```
gwatts: was there a reason to make this a class, rather than just decorating a function with `@cozmo.annotate.annotator`  ?

examples/remote_control_cozmo.py
```diff
@@ -32,11 +32,41 @@
...
+        d = ImageDraw.Draw(image)
+
+        bounds = [3, 0, image.width, image.height]
+
+        def print_line(text_line):
+            nonlocal bounds
+            text = cozmo.annotate.ImageText(text_line, position=cozmo.annotate.TOP_LEFT, color='lightblue')
+            text.render(d, bounds)
+            TEXT_HEIGHT = 11
+            bounds[1] += TEXT_HEIGHT
```
gwatts: The `text` argument to`ImageText` should work ok with a multi-line string so not sure this `print_line` function is necessary

examples/remote_control_cozmo.py
MarkWesleyAtAnki: No, I just followed the pattern that Battery uses in annotate.py. Is there an advantage of either approach - they look to be almost identical in terms of amount of code to type?

gwatts: ah.. nope; end result is the same - just less boilerplate & indentation with the decorator

examples/remote_control_cozmo.py
MarkWesleyAtAnki: ah, yes, I'd needed nonlocal when i was using a tuple and having to do bounds = (bounds[0], ...) each time but can remove this now.

examples/remote_control_cozmo.py
MarkWesleyAtAnki: I didn't realize that ImageText supported multi-line, although it doesn't give control of line-spacing (and spaces too far apart), and building a multiline string with "\n" separators feels uglier to me.

examples/remote_control_cozmo.py
```diff
@@ -32,11 +32,41 @@
...
+
+        pose = robot.pose
+        print_line('Pose: Pos = <%.1f, %.1f, %.1f>' % pose.position.x_y_z)
+        print_line('Pose: Rot quat = <%.1f, %.1f, %.1f, %.1f>' % pose.rotation.q0_q1_q2_q3)
+        print_line('Pose: angle_z = %.1f' % pose.rotation.angle_z.degrees)
+        print_line('Pose: origin_id: %s' % pose.origin_id)
+
+        # Display the Accelerometer and Gyro data for the robot
+
+        print_line('Acclrm: <%.1f, %.1f, %.1f>' % robot.accelerometer.x_y_z)
```
msintov: Could 'Acclrm' be expanded anymore to be more comprehensible?

examples/remote_control_cozmo.py
```diff
@@ -58,7 +88,11 @@ def create_default_image(image_width, image_height, do_gradient=False):
 remote_control_cozmo = None
 _default_camera_image = create_default_image(320, 240)
 _is_mouse_look_enabled_by_default = False
-_display_debug_annotations = True
+
+DEBUG_ANNOTATIONS_DISABLED = 0
```
msintov: I would expect these 3 constants to be at the top of the script near the imports instead.

src/cozmo/robot.py
```diff
@@ -625,7 +638,9 @@ def anim_names(self):
 
     @property
     def pose(self):
-        """:class:`cozmo.util.Pose`: The current pose of cozmo relative to where he started when the engine was initialized.
+        """:class:`cozmo.util.Pose`: The current pose of cozmo relative to where it started
```
msintov: Cozmo should be uppercase in "of cozmo".

examples/remote_control_cozmo.py
MarkWesleyAtAnki: I don't have time to add that, I don't see the issue with using a print_line function (which provides more flexibility for e.g. coloring or indentation), but I'll switch it to building a "\n" separated block and printing that at the cost of an uglier display.

gwatts: no leave it as is - it's not that it's bad - it's that i'd rather improve the SDK than introduce hacks into examples.

If there's no time now then I'd just put it on the list for later

examples/remote_control_cozmo.py
MarkWesleyAtAnki: "Accelerometer" takes up a lot of space in the image, and I was worried that "Acc" might be confusing (as in "why does it show Cozmo accelerating when he's stationary", and apparently "Acclrm" is the abbreviation for that, but i agree it looks awkward. I'll go with "Accel"?

PR #39, opened 2016-11-30
COZMO-7518 Add IR Headlight control

MarkWesleyAtAnki:
Add robot.set_head_light method to turn IR headlight on and off
Added keyboard and button mappings to turn IR light on and off in the remote_control_cozmo example program

### DIFF
--- a/examples/remote_control_cozmo.py
+++ b/examples/remote_control_cozmo.py
@@ -444,6 +444,7 @@ def handle_index_page():
                         <h3>General:</h3>
                         <b>Shift</b> : Hold to Move Faster (Driving, Head and Lift)<br>
                         <b>Alt</b> : Hold to Move Slower (Driving, Head and Lift)<br>
+                        <b>L</b> : Toggle IR HeadLight: <button id="headLightId" onClick=onHeadLightButtonClicked(this) style="font-size: 14px">Default</button><br>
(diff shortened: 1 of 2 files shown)


examples/remote_control_cozmo.py
```diff
@@ -444,6 +444,7 @@ def handle_index_page():
                         <h3>General:</h3>
                         <b>Shift</b> : Hold to Move Faster (Driving, Head and Lift)<br>
                         <b>Alt</b> : Hold to Move Slower (Driving, Head and Lift)<br>
+                        <b>L</b> : Toggle IR HeadLight: <button id="headLightId" onClick=onHeadLightButtonClicked(this) style="font-size: 14px">Default</button><br>
```
msintov: "HeadLight" does not typically have an internal uppercase L (so normally it is "Headlight"). Can we make the internal uppercase L be lowercase throughout or at least everywhere the user will see it (i.e., text, code)?

examples/remote_control_cozmo.py
```diff
@@ -679,6 +695,15 @@ def handle_setMouseLookEnabled():
     return ""
 
 
+@flask_app.route('/setHeadLightEnabled', methods=['POST'])
+def handle_setHeadLightEnabled():
+    '''Called from Javascript whenever head-light enabled is toggled'''
```
msintov: "head-light" should be "headlight"

I think this is not quite precise: "Called from Javascript whenever head-light enabled is toggled". Maybe instead: "Called from Javascript when headlight is toggled" or "when headlight mode is toggled"?

PR #42, opened 2016-12-05
COZMO-7567 Get Available Objects before Robot Ready

MarkWesleyAtAnki:
Request the available objects earlier, before requesting the available animations - this ensures we have received the available objects before we mark the robot as ready, because we already wait on the animation list to arrive. (Note: adding an additional check to wait for available objects wouldn't be possible without an app change, because there's no guarantee of any message in the case that all cubes and charger are disconnected).

### DIFF
--- a/src/cozmo/conn.py
+++ b/src/cozmo/conn.py
@@ -376,10 +376,18 @@ class CozmoConnection(event.Dispatcher, clad_protocol.CLADProtocol):
 
         self._is_ui_connected = True
         self.dispatch_event(EvtConnected, conn=self)
-        self.anim_names.refresh()
+
         logger.info('App connection established. sdk_version=%s '
                 'cozmoclad_version=%s app_build_version=%s',
                 version.__version__, cozmoclad.__version__, msg.buildVersion)
+
+        # We RequestAvailableObjects before the animation names as this ensures
+        # that we will receive the responses before we mark the robot as ready
+        msg = _clad_to_engine_iface.RequestAvailableObjects()
+        self.send_msg(msg)
+
+        self.anim_names.refresh()
+
 
     def _recv_msg_image_chunk(self, evt, *, msg):
         if self._primary_robot:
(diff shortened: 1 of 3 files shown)


src/cozmo/conn.py
```diff
@@ -376,11 +376,19 @@ def _handle_ui_device_connected(self, msg):
 
         self._is_ui_connected = True
         self.dispatch_event(EvtConnected, conn=self)
-        self.anim_names.refresh()
+
         logger.info('App connection established. sdk_version=%s '
                 'cozmoclad_version=%s app_build_version=%s',
                 version.__version__, cozmoclad.__version__, msg.buildVersion)
 
+        # We RequestAvailableObjects before the animation names as this ensures
```
msintov: Instead of "We RequestAvailableObjects before the animation names", how about "We call RequestAvailableObjects before refreshing the animation names"?

src/cozmo/conn.py
```diff
@@ -376,11 +376,19 @@ def _handle_ui_device_connected(self, msg):
...
+        # that we will receive the responses before we mark the robot as ready
+        msg = _clad_to_engine_iface.RequestAvailableObjects()
+        self.send_msg(msg)
+
+        self.anim_names.refresh()
+
```
msintov: Nitpick: there's an extra newline here after the method

src/cozmo/objects.py
```diff
@@ -348,8 +348,11 @@ def _recv_msg_robot_observed_object(self, evt, *, msg):
     #### Commands ####
 
 
+#: LightCube1Id's markers look a bit like a paperclip
 LightCube1Id = _clad_to_game_cozmo.ObjectType.Block_LIGHTCUBE1
+#: LightCube2Id's markers look a bit like a lamp
 LightCube2Id = _clad_to_game_cozmo.ObjectType.Block_LIGHTCUBE2
+#: LightCube3Id's markers look a bit like the letters ab over T
```
msintov: Maybe change "like the letters ab over T" to "like the letters 'ab' over 'T'"

PR #43, opened 2016-12-05
COZMO-7699 Show Facial Features

MarkWesleyAtAnki:
Now that facial features are available, store them in the face and draw them in the debug annotations
Allow facial expression estimation to be enabled and tracked, and draw it on annotations when relevant
Fix PET_TYPE_UNKOWN typo to PET_TYPE_UNKNOWN

### DIFF
--- a/src/cozmo/annotate.py
+++ b/src/cozmo/annotate.py
@@ -31,7 +31,7 @@ __all__ = ['DEFAULT_OBJECT_COLORS',
            'TOP_LEFT', 'TOP_RIGHT', 'BOTTOM_LEFT', 'BOTTOM_RIGHT',
            'ImageText', 'Annotator', 'ObjectAnnotator', 'FaceAnnotator',
            'PetAnnotator', 'TextAnnotator', 'ImageAnnotator',
-           'annotator', 'add_img_box_to_image', 'add_polygon_to_image']
+           'add_img_box_to_image', 'add_polygon_to_image', 'annotator']
 
 
 import collections
@@ -149,17 +149,17 @@ def add_img_box_to_image(image, box, color, text=None):
 def add_polygon_to_image(image, poly_points, scale, line_color, fill_color=None):
     '''Draw a polygon on an image
 
-    This will draw a polygon on the passed in image in the specified
+    This will draw a polygon on the passed-in image in the specified
     colors and scale.
 
     Args:
         image (:class:`PIL.Image.Image`): The image to draw on
-        poly_points: A sequence of points representing the the polygon,
+        poly_points: A sequence of points representing the polygon,
             where each point has float members (x, y)
-        scale (float): Scale to multiply each point by to match the image scaling
-        line_color (string): The color for the outline of the polygon, string
+        scale (float): Scale to multiply each point to match the image scaling
+        line_color (string): The color for the outline of the polygon. The string value
             must be a color string suitable for use with PIL - see :mod:`PIL.ImageColor`
-        fill_color (string): The color for the inside of the polygon, string
+        fill_color (string): The color for the inside of the polygon. The string value
             must be a color string suitable for use with PIL - see :mod:`PIL.ImageColor`
     '''
     if len(poly_points) < 2:
--- a/src/cozmo/annotate.py
+++ b/src/cozmo/annotate.py
@@ -31,7 +31,7 @@ __all__ = ['DEFAULT_OBJECT_COLORS',
            'TOP_LEFT', 'TOP_RIGHT', 'BOTTOM_LEFT', 'BOTTOM_RIGHT',
            'ImageText', 'Annotator', 'ObjectAnnotator', 'FaceAnnotator',
            'PetAnnotator', 'TextAnnotator', 'ImageAnnotator',
-           'annotator', 'add_img_box_to_image']
+           'annotator', 'add_img_box_to_image', 'add_polygon_to_image']
 
 
 import collections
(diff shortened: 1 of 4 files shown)


src/cozmo/annotate.py
```diff
@@ -31,7 +31,7 @@
            'TOP_LEFT', 'TOP_RIGHT', 'BOTTOM_LEFT', 'BOTTOM_RIGHT',
            'ImageText', 'Annotator', 'ObjectAnnotator', 'FaceAnnotator',
            'PetAnnotator', 'TextAnnotator', 'ImageAnnotator',
-           'annotator', 'add_img_box_to_image']
+           'annotator', 'add_img_box_to_image', 'add_polygon_to_image']
```
msintov: For alphabetization, should annotator follow the two add*?

src/cozmo/annotate.py
```diff
@@ -146,6 +146,35 @@ def add_img_box_to_image(image, box, color, text=None):
             text.render(d, (x1, y1, x2, y2))
 
 
+def add_polygon_to_image(image, poly_points, scale, line_color, fill_color=None):
+    '''Draw a polygon on an image
+
+    This will draw a polygon on the passed in image in the specified
```
msintov: "passed in" should be "passed-in"

src/cozmo/annotate.py
```diff
@@ -146,6 +146,35 @@ def add_img_box_to_image(image, box, color, text=None):
...
+    colors and scale.
+
+    Args:
+        image (:class:`PIL.Image.Image`): The image to draw on
+        poly_points: A sequence of points representing the the polygon,
```
msintov: Should remove one 'the' from: "representing the the polygon"

src/cozmo/annotate.py
```diff
@@ -146,6 +146,35 @@ def add_img_box_to_image(image, box, color, text=None):
...
+            where each point has float members (x, y)
+        scale (float): Scale to multiply each point by to match the image scaling
```
msintov: Regardining "Scale to multiply each point by to match the image scaling"

Should we remove the 'by' so that it is: "Scale to multiply each point to match the image scaling"?

src/cozmo/annotate.py
```diff
@@ -146,6 +146,35 @@ def add_img_box_to_image(image, box, color, text=None):
...
+        line_color (string): The color for the outline of the polygon, string
```
msintov: Seems like this should be two sentences, the first ending with "polygon." Then the second sentence starting with something like "The string value...".

src/cozmo/annotate.py
```diff
@@ -146,6 +146,35 @@ def add_img_box_to_image(image, box, color, text=None):
...
+            must be a color string suitable for use with PIL - see :mod:`PIL.ImageColor`
+        fill_color (string): The color for the inside of the polygon, string
```
msintov: Same two sentence comment as above

src/cozmo/annotate.py
```diff
@@ -238,15 +267,22 @@ def apply(self, image, scale):
...
 
         Override or replace to customize.
         '''
+        expression = obj.known_expression
+        if len(expression) > 0:
+            expression += " "
         if obj.name:
-            return ImageText(obj.name)
-        return ImageText('(unknown face %d)' % obj.face_id)
+            return ImageText('%s%s (%d)' % (expression, obj.name, obj.face_id))
```
msintov: Should there be a space between the two %s format specifiers?

src/cozmo/annotate.py
```diff
@@ -238,15 +267,22 @@ def apply(self, image, scale):
...
+        return ImageText('(unknown%s face %d)' % (expression, obj.face_id))
```
msintov: Should there be a space after 'unknown' and before %s?

src/cozmo/faces.py
```diff
@@ -262,14 +302,68 @@ def name(self):
...
+        `FACIAL_EXPRESSION_SURPRISED`, `FACIAL_EXPRESSION_ANGRY`,
+        or `FACIAL_EXPRESSION_SAD`.
+        '''
+        return self._expression
+
+    @property
+    def known_expression(self):
+        '''string: The known facial expression Cozmo has recognized on the face.
+
+        Like :meth:`expression` but returns an empty string for the unknown expression
```
msintov: Seems like this should end with a period ('...unknown expression.').

src/cozmo/pets.py
```diff
@@ -33,7 +33,7 @@
 '''
 
 # __all__ should order by constants, event classes, other classes, functions.
-__all__ = ['PET_VISIBILITY_TIMEOUT', 'PET_TYPE_UNKOWN', 'PET_TYPE_CAT', 'PET_TYPE_DOG',
+__all__ = ['PET_VISIBILITY_TIMEOUT', 'PET_TYPE_UNKNOWN', 'PET_TYPE_CAT', 'PET_TYPE_DOG',
```
msintov: The constants should be in alpha order starting with 'PET_TYPE_CAT', 'PET_TYPE_DOG', etc.

src/cozmo/annotate.py
MarkWesleyAtAnki: no, that's deliberate - if the string is empty (e.g. face expression estimation is off) then we don't want to add a stray space in

src/cozmo/annotate.py
MarkWesleyAtAnki: no, that's deliberate - if the string is empty (e.g. face expression estimation is off) then we don't want to add a stray space in

PR #49, opened 2016-12-06
Kevin/cozmo 7697 rename available objects

kevinatanki:
This includes almost-latest master, which I assume is ok. My changes are just in the one commit though to support a renamed message.


MarkWesleyAtAnki:
@kevinatanki I'd prefer it if the master->nextclad merge was it's own PR, and this was just the AvailableObjects change

PR #52, opened 2016-12-08
Support ActionResult changes

alchaussee:
Clad changes made by https://github.com/anki/cozmo-one/pull/2881
Can search for "clad/src" on the "Files changed" tab to see what changed

### DIFF
--- a/src/cozmo/action.py
+++ b/src/cozmo/action.py
@@ -210,16 +210,15 @@ class Action(event.Dispatcher):
         elif result == types.INTERRUPTED:
             self._set_failed('interrupted', 'Action was interrupted')
 
-        # Can add additional if statements for other ActionResults to exposure more specific failure reasons to users
-
         else:
-            # Got some other result figure out if it is a result that indicates the action can be retried
-            resultCategory = result >> _clad_to_game_cozmo.ARCBitShift.NUM_BITS
-            resultCategories = _clad_to_game_cozmo.ActionResultCategory
+            # All other results should fall under either the abort or retry
+            # categories, determine the category by shifting the result
+            result_category = result >> _clad_to_game_cozmo.ARCBitShift.NUM_BITS
+            result_categories = _clad_to_game_cozmo.ActionResultCategory
             
-            if resultCategory == resultCategories.ABORT:
+            if result_category == result_categories.ABORT:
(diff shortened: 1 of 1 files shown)


src/cozmo/action.py
```diff
@@ -219,9 +210,21 @@ def _recv_msg_robot_completed_action(self, evt, *, msg):
         elif result == types.INTERRUPTED:
             self._set_failed('interrupted', 'Action was interrupted')
 
+        # Can add additional if statements for other ActionResults to exposure more specific failure reasons to users
```
MarkWesleyAtAnki: typo: exposure should be expose. We also try and wrap at about 80 chars for the SDK. I think we should just remove this comment and add some more detail in the else block though (about how we're summarizing the types by their category).

src/cozmo/action.py
```diff
@@ -219,9 +210,21 @@ def _recv_msg_robot_completed_action(self, evt, *, msg):
...
+
         else:
-            self._set_failed('unknown', 'Action failed with unknown reason')
-            logger.error('Received unknown action result status %s', msg)
+            # Got some other result figure out if it is a result that indicates the action can be retried
```
MarkWesleyAtAnki: We try and wrap at about 80 chars for the SDK, suggest rewording to:
```
            # All other results should fall under either the abort or retry
            # categories, determine the category by shifting the result.
```

src/cozmo/action.py
```diff
@@ -219,9 +210,21 @@ def _recv_msg_robot_completed_action(self, evt, *, msg):
...
+            resultCategory = result >> _clad_to_game_cozmo.ARCBitShift.NUM_BITS
+            resultCategories = _clad_to_game_cozmo.ActionResultCategory
+            
+            if resultCategory == resultCategories.ABORT:
+                self._set_failed('aborted', 'Action failed')
```
MarkWesleyAtAnki: Ideally we would add the result to the message, so e.g. 'Action failed: NO_FACE', but we don't have an automatic enum->string conversion in the python clad generator currently, so you have to build something each time - we can add this later though, so don't worry about it for now (this is mostly just a note for myself).

alchaussee: Yeah we should definitely do that

PR #59, opened 2016-12-15
COZMO-7650 SDK webots test

lip:
*  Updating some clad messages
 *  Exposing the TCPConnector class for use to connect to webots

### DIFF
--- a/src/cozmo/action.py
+++ b/src/cozmo/action.py
@@ -189,17 +189,17 @@ class Action(event.Dispatcher):
             logger.warn('Received "running" action notification for action=%s', self)
             self._set_failed('running', 'Action was still running')
 
-        elif result == types.NOT_STARTED:
+        elif result == types.FAILURE_NOT_STARTED:
             # not sure we'll see this?
             self._set_failed('not_started', 'Action was not started')
 
-        elif result == types.TIMEOUT:
+        elif result == types.FAILURE_TIMEOUT:
             self._set_failed('timeout', 'Action timed out')
 
-        elif result == types.TRACKS_LOCKED:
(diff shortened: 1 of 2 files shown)


lip:
Is there any concern with exposing the TCPConnector class?

MarkWesleyAtAnki:
No, I think exposing the TCPConnector class is fine.

src/cozmo/action.py
```diff
@@ -189,17 +189,17 @@ def _recv_msg_robot_completed_action(self, evt, *, msg):
             logger.warn('Received "running" action notification for action=%s', self)
             self._set_failed('running', 'Action was still running')
 
-        elif result == types.NOT_STARTED:
+        elif result == types.FAILURE_NOT_STARTED:
```
MarkWesleyAtAnki: I think the naming on the left (without FAILURE_) was to match the latest code in cozmo-one master, or did that change back again? (python-sdk master is still like the change on the right).

lip: Looks like it changed back for what ever reason
https://github.com/anki/cozmo-one/blob/master/clad/src/clad/types/actionTypes.clad

MarkWesleyAtAnki: Hmmm, looks like @alchaussee 's matching change for cozmo-one didn't go in? In which case there will also be runtime python issues on aborting or failing actions - how far off is that cozmo-one change from going in (these should really go in at the same time - ultimately we'll probably need to use a sub-module pointer into specific commits in nextclad), otherwise we should probably revert that earlier commit and re-apply it once it's ready.

lip:
I was not aware that the action result change was still pending.  I ended up reverting it

PR #62, opened 2016-12-20
COZMO-7694 Check validity of poses

MarkWesleyAtAnki:
When receiving object info on connection - poses where PoseState is unknown are always invalid and should just be ignored.
Added support to Util.Pose for tracking validity.

### DIFF
--- a/src/cozmo/util.py
+++ b/src/cozmo/util.py
@@ -388,7 +388,7 @@ class Pose:
     @property
     def is_valid(self):
         '''bool: Returns True if this is a valid, usable pose.'''
-        return self.origin_id() >= 0
+        return self.origin_id >= 0
 
     @property
     def position(self):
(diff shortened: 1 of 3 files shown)


src/cozmo/util.py
```diff
@@ -368,6 +368,28 @@ def encode_pose(self):
...
+        Returns:
+            bool: True if the two poses are comparable, False otherwise.
+        '''
+        return (self.is_valid and other_pose.is_valid and
+                (self.origin_id == other_pose.origin_id))
+
+    @property
+    def is_valid(self):
+        '''bool: Returns True if this is a valid, usable pose.'''
+        return self.origin_id() >= 0
```
gwatts: shouldn't this be `return self.origin_id >= 0` as origin_id is a property?

MarkWesleyAtAnki:
Merged via https://github.com/anki/cozmo-python-sdk/pull/65

PR #60, opened 2016-12-17
COZMO-7707 Improve backpack and animation documentation

MarkWesleyAtAnki:
Documented what each of the 5 backpack lights refers to.
Added not on backpack lights regarding left/right being red LED only
Added set_center_backpack_lights function that forces left/right LEDs to off, which is the more likely use-case.
Added warning to play_anim regarding how we recommend play_anim_trigger instead as individual animations may be renamed or removed in future app versions.

PR #63, opened 2016-12-21
Documented what each of the 5 backpack lights refers to.

MarkWesleyAtAnki:
Added note on backpack lights regarding left/right being red LED only
Added set_center_backpack_lights function that forces left/right LEDs to off, which is the more likely use-case.
Added warning to play_anim regarding how we recommend play_anim_trigger instead as individual animations may be renamed or removed in future app versions.


MarkWesleyAtAnki:
Reviewed in https://github.com/anki/cozmo-python-sdk/pull/60
had to create a clean squashed branch to actually merge it

PR #61, opened 2016-12-18
COZMO-6362 Add support for running actions in parallel

MarkWesleyAtAnki:
Adding in_parallel parameter for all actions, this tells engine to run in parallel (and not clear current actions), and also prevents the SDK from raising an exception if there were pending actions in that case.
OLED face image is now an sent as a parallel action, with previous face actions automatically aborted each time
Aborted actions are immediately not considered in-progress (no longer have to wait a frame for engine to return the cancelled message before starting another action)

PR #64, opened 2016-12-21
COZMO-6362 Parallel Action Support

MarkWesleyAtAnki:
Adding in_parallel parameter for all actions, this tells engine to run in parallel (and not clear current actions), and also prevents the SDK from raising an exception if there were pending actions in that case.
OLED face image is now an sent as a parallel action, with previous face actions automatically aborted each time
Aborted actions are immediately not considered in-progress (no longer have to wait a frame for engine to return the cancelled message before starting another action)
Added 06_actions/01_parallel_actions.py example to demonstrate the above.

### DIFF
--- a/examples/tutorials/06_actions/01_parallel_actions.py
+++ b/examples/tutorials/06_actions/01_parallel_actions.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2016 Anki, Inc.
+#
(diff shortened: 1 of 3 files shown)


MarkWesleyAtAnki:
Reviewed in https://github.com/anki/cozmo-python-sdk/pull/61
Had to squash into a new branch to merge into master cleanly.

PR #65, opened 2016-12-21
COZMO-7694 Check validity of poses

MarkWesleyAtAnki:
When receiving object info on connection - poses where PoseState is unknown are always invalid and should just be ignored.
Added support to Util.Pose for tracking validity.
Updated 08_drive_to_charger_test to show usage.


MarkWesleyAtAnki:
Reviewed in https://github.com/anki/cozmo-python-sdk/pull/62
Had to squash to a clean branch to merge cleanly into master

PR #67, opened 2016-12-21
Update Makefile for new example layout

gwatts:
Also emove real target phony dependencies to ensure targets are only built
when necessary

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: copy-clad license sdist wheel examples dist
+.PHONY: copy-clad dist examples license wheel vagrant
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: copy-clad license sdist wheel examples dist
+.PHONY: copy-clad dist excamples license wheel vagrant
```
MarkWesleyAtAnki: is excamples a typo ?

MarkWesleyAtAnki:
Also looks like a typo in the commit message "Also emove real"

Makefile
gwatts: ick yes

gwatts:
fixed the commit message typo

PR #69, opened 2017-01-09
Improve Windows adb instructions

msintov:
Simplify the Windows adb install instructions by using a new link that downloads platform-tools directly instead of using the Android Studio command line tools download and installer. Google promises to keep the latest platform-tools available at the new link.

With the new instructions the user no longer needs to run the android/SDK Manager installer tool, and because of this, the user also no longer needs to install Java.

### DIFF
--- a/docs/source/adb.rst
+++ b/docs/source/adb.rst
@@ -22,32 +22,23 @@ macOS Installation
 Windows Installation
 ^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


msintov:
>Is Java no longer required then (was that just for the installer program?!)

Correct, Java is no longer required. At least, I don't see Java installed on my Windows WM, running 'java' at the command line does not work, and adb still recognizes my devices.

PR #71, opened 2017-01-12
Fix range check for with_alpha

touretzkyds:
with_alpha rejects alpha values of 0 or 255 due to a faulty range check.


MarkWesleyAtAnki:
@touretzkyds
We've finalized our CLA process that will allow us to accept contributions from 3rd parties. We need anyone contributing to first complete an individual CLA:
http://go.anki.com/individual-cla
If your employer(s) have rights to intellectual property that you create that includes your Contributions, then we also require a separate Corporate CLA from them:
http://go.anki.com/corporate-cla

MarkWesleyAtAnki:
The (unused) alpha setting from lights has been removed, this PR is no longer relevant.

PR #78, opened 2017-02-23
COZMO-9686 Fix rare exception from aborting a just-completed action

MarkWesleyAtAnki:
Aborting an action that just completed (but didn’t yet receive the event to update from running to complete) no longer results in an exception but now gracefully does nothing.
AnimationTrigger class now correctly caches and returns completed event in the case of WaitForCompleted() being called when already complete
Animation class now behaves like AnimationTrigger in dispatching EvtAnimationCompleted on completion instead of base Action’s event

### DIFF
--- a/src/cozmo/action.py
+++ b/src/cozmo/action.py
@@ -105,6 +105,7 @@ class Action(event.Dispatcher):
         self._failure_code = None
         self._failure_reason = None
         self._completed_event = None
+        self._completed_event_pending = False
 
     def __repr__(self):
         extra = self._repr_values()
@@ -126,6 +127,7 @@ class Action(event.Dispatcher):
 
     def _set_completed(self, msg):
         self._state = ACTION_SUCCEEDED
+        self._completed_event_pending = False
         self._dispatch_completed_event(msg)
 
     def _dispatch_completed_event(self, msg):
@@ -139,6 +141,7 @@ class Action(event.Dispatcher):
         self._state = ACTION_FAILED
         self._failure_code = code
         self._failure_reason = reason
+        self._completed_event_pending = False
         self._completed_event = EvtActionCompleted(action=self, state=self._state,
                                               failure_code=code,
                                               failure_reason=reason)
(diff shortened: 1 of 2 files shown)


src/cozmo/action.py
```diff
@@ -442,13 +448,22 @@ def _abort_action(self, action):
...
-        self.robot.conn.send_msg(msg)
+        if action._completed_event_pending:
+            # The action was marked as still running but the ActionDispatcher
+            # has already received a completion message (and removed it from
+            # _in_progress) - the action is just waiting to receive a
+            # robot_completed_action message that is still being dispatched
+            # via asyncio.ensure_future
+            logger.debug('Not sending abort for action=%s to engine as it just completed', action)
+        else:
+            # move from in-progress to aborting dicts
```
msintov: I'm curious why this else case code is new and wasn't needed prior to this change (moving the _action_id from _aborting to _in_progress)?

src/cozmo/anim.py
```diff
@@ -89,10 +95,10 @@ def _encode(self):
             robotID=self.robot.robot_id, trigger=self.trigger.id, numLoops=self.loop_count)
 
     def _dispatch_completed_event(self, msg):
-        self.dispatch_event(EvtAnimationCompleted,
+        self._completed_event = EvtAnimationCompleted(
```
msintov: With the change to this method, how is the behavior different than it was before?

src/cozmo/action.py
MarkWesleyAtAnki: The else case isn't new, it's just indented, but github's diff isn't making that clear - you can add `?w=1` on the end of github diff urls to ignore whitespace which is handy sometimes - e.g: https://github.com/anki/cozmo-python-sdk/pull/78/files/d534520b8a575284cc129f479148b5289655730a?w=1

src/cozmo/anim.py
MarkWesleyAtAnki: The completed event is now stored in `self._completed_event` - this is done for all other Action types, but this one overrode `_dispatch_completed_event` and wasn't doing it. Prior to this change if you did an AnimationTrigger action, slept long enough for it to complete, and then called action.WaitForCompleted() you would receive a None object instead of the CompletedEvent which could potentially have broken client code if they were using that data for anything.

PR #79, opened 2017-02-23
For 1.3 Don't Merge COZMO-9149 Color images

alchaussee:
Code to support https://github.com/anki/cozmo-one/pull/3378


MarkWesleyAtAnki:
@alchaussee will this require 1.3 - if so we'll need another nextclad (as currently nextclad is for stuff in 1.2 until we ship 1.2)

alchaussee:
Oh right... I will update the title of this PR as don't merge until 1.3 or something like that and as long as you don't approve it, I can't merge it. Or I could just close it and reopen it at a later date

src/cozmo/camera.py
```diff
@@ -244,12 +280,62 @@ def _minigray_to_jpeg(minigray,  width, height):
...
+             0x02, 0x00, # Cb same AC/DC
+             0x03, 0x00, # Cr same AC/DC
+             0x00, 0x3F, 0x00
+            ], dtype=np.uint8)
+
+        return _mini_to_jpeg_helper(minicolor, width, height, header)
+
+@_require_img_processing
+def _mini_to_jpeg_helper(mini, width, height, header):
+        # Does not work correctly yet
```
MarkWesleyAtAnki: Is this "Does not work correctly yet" comment still valid (I think it's old and should have been removed a while ago, as far as I'm aware this all works now).

PR #81, opened 2017-02-28
COZMO-7787 Expose the result of an Action more clearly.

MarkWesleyAtAnki:
Wrapped the ActionResult from CLAD with a documented Python class to more clearly explain the meaning of each result.
Corrected the action documentation to account for in_parallel usage, and how ActionResult relates to that.
Updated cube_stack example to print the result for failed pickup or place actions.

### DIFF
--- a/src/cozmo/action.py
+++ b/src/cozmo/action.py
@@ -116,7 +116,7 @@ class ActionResults:
     #: There was a problem with the Pose provided.
     BAD_POSE = _ActionResult("BAD_POSE", _clad_to_game_cozmo.ActionResult.BAD_POSE)
 
-    # (Undocumented) The SDK provided tag was bad (shouldn't occur - would indicate a bug in the SDK)
+    # (Undocumented) The SDK-provided tag was bad (shouldn't occur - would indicate a bug in the SDK)
     BAD_TAG = _ActionResult("BAD_TAG", _clad_to_game_cozmo.ActionResult.BAD_TAG)
 
     # (Undocumented) Shouldn't occur outside of factory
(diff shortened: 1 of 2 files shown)


src/cozmo/action.py
```diff
@@ -65,6 +74,212 @@
...
+    #: There was an error related to vision markers.
+    BAD_MARKER = _ActionResult("BAD_MARKER", _clad_to_game_cozmo.ActionResult.BAD_MARKER)
+
+    #: There was a problem with the Object ID provided (e.g. there is no Object with that ID).
+    BAD_OBJECT = _ActionResult("BAD_OBJECT", _clad_to_game_cozmo.ActionResult.BAD_OBJECT)
+
+    #: There was a problem with the Pose provided.
+    BAD_POSE = _ActionResult("BAD_POSE", _clad_to_game_cozmo.ActionResult.BAD_POSE)
+
+    # (Undocumented) The SDK provided tag was bad (shouldn't occur - would indicate a bug in the SDK)
```
msintov: What does "Undocumented" mean in the various new places it appears?

src/cozmo/action.py
```diff
@@ -65,6 +74,212 @@
...
+
+    #: Failed to follow the planned path.
+    FAILED_TRAVERSING_PATH = _ActionResult("FAILED_TRAVERSING_PATH",
+                                           _clad_to_game_cozmo.ActionResult.FAILED_TRAVERSING_PATH)
+
+    #: The previous attempt to pick and place an object failed.
+    LAST_PICK_AND_PLACE_FAILED = _ActionResult("LAST_PICK_AND_PLACE_FAILED",
+                                               _clad_to_game_cozmo.ActionResult.LAST_PICK_AND_PLACE_FAILED)
+
+    #: The required motor isn't movign so the action cannot complete.
```
msintov: "movign" misspelling

src/cozmo/action.py
```diff
@@ -65,6 +74,212 @@
...
+    MOTOR_STOPPED_MAKING_PROGRESS = _ActionResult("MOTOR_STOPPED_MAKING_PROGRESS",
+                                                _clad_to_game_cozmo.ActionResult.MOTOR_STOPPED_MAKING_PROGRESS)
+
+    #: Not carrying an object when it was expected, but may suceed if the action is retried.
```
msintov: s/b "succeed"

src/cozmo/action.py
```diff
@@ -65,6 +74,212 @@
...
+    NOT_CARRYING_OBJECT_RETRY = _ActionResult("NOT_CARRYING_OBJECT_RETRY",
+                                              _clad_to_game_cozmo.ActionResult.NOT_CARRYING_OBJECT_RETRY)
+
+    #: Cozmo was unable to plan a path, but may suceed if the action is retried.
```
msintov: "succeed"

src/cozmo/action.py
MarkWesleyAtAnki: A number of these results only make sense for internal dev (or factory test/calibration), so I felt it best to keep them out of the API documentation to avoid confusion. The thing that's keeping them out of the API docs is just that it's commented with just a hash `#` instead of a hash-colon `#:` so it won't get a docstring - I just added the "(Undocumented)" text to help clarify that it was deliberately excluded. I guess it's not that clear on reflection...

PR #84, opened 2017-02-28
COZMO-9810 Turn off cube lights on SDK connection

MarkWesleyAtAnki:
A bug was introduced on the app-side where the EnableLightStates(enable=False) message (which is automatically sent to disable the cube lights on entering SDK mode and at the end of every SDK program) no longer turns off the lights as part of changing the states. This will be resolved for the 1.3 app release, but in the meantime we’ll automatically turn the lights off on cubes when SDK first connects to them each time to avoid cube lights persisting across multiple SDK program runs.

### DIFF
--- a/src/cozmo/world.py
+++ b/src/cozmo/world.py
@@ -155,10 +155,16 @@ class World(event.Dispatcher):
             if not cube:
                 logger.error('Received invalid cube objecttype=%s msg=%s', msg.objectType, msg)
                 return
+            is_uninitialized_cube = (cube.object_id == None)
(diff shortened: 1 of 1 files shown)


msintov:
Is it possible to also turn the lights off at the end of the SDK program for this PR/FIXME?

MarkWesleyAtAnki:
There's no way to guarantee that something would be sent from SDK at the end (especially in the event of connection loss), and I don't think it's worth spending extra time to slightly improve this band-aid when we'll get rid of it on the next release anyway.

PR #85, opened 2017-03-01
COZMO-9710 Add Youtube installation videos into the documentation

MarkWesleyAtAnki:
Embedded the 2 mac videos in the macOS page, and the 2 Windows ones to the Windows installation page.
Embedded the basic walkthrough in the get started section
Fixed warning from adb.rst generation (you can’t have 2 links with the same name unless you make them anonymous with a double trailing underscore instead of a single one).

### DIFF
--- a/docs/source/getstarted.rst
+++ b/docs/source/getstarted.rst
@@ -38,6 +38,18 @@ Starting Up the SDK
 Example Programs
 ----------------
 
+^^^^^^^^^^^^^^^^^
+Walkthrough Video
+^^^^^^^^^^^^^^^^^
+
+For your convenience, here is a video detailing the first 8 simple example programs. There is also simple text-based walkthrough of running your first program below that.
+
+.. raw:: html
+
+   <iframe width="690" height="388" src="https://www.youtube.com/embed/YAQ_USpkxgE?rel=0" frameborder="0" allowfullscreen></iframe>
+
+|
+
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 First Steps - "Hello World"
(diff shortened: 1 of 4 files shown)


docs/source/getstarted.rst
```diff
@@ -38,6 +38,18 @@ Starting Up the SDK
 Example Programs
 ----------------
 
+^^^^^^^^^^^^^^^^^
+Walkthrough Video
+^^^^^^^^^^^^^^^^^
+
+For your convenience, here is a video detailing the first 8 simple example programs. There is also simple text-based walkthrough of running your first program below that.
```
msintov: Maybe instead of "There is also simple text-based walkthrough of running your first program below that", you could say, "You can also find a simple text-based walkthrough of running your first program below" or similar.

Maybe writing 8 here isn't future proof? Don't want to have to remember to come back here. Or maybe it's fine.

docs/source/install-macos.rst
```diff
@@ -9,6 +9,20 @@ Installation - macOS / OS X
 This guide provides instructions on installing the Cozmo SDK for computers running with a macOS operating system.
 
 ^^^^^^^^^^^^^^^^^^^
+Installation Videos
+^^^^^^^^^^^^^^^^^^^
+
+For your convenience, videos are provided showing the installation steps being followed on a macOS / OS X computer. One using an iOS device, and one using an Android device. There is also full text-based documenation below these.
```
msintov: Correct spelling: "documentation"

The sentence "One using an iOS device, and one using an Android device", should be part of the previous sentence, and shouldn't be a sentence on its own.

PR #89, opened 2017-03-15
Mwesley/cozmo 4610 pose state refactor

MarkWesleyAtAnki:
Split ObjectStates into ConnectedObjectStates and LocatedObjectStates
Split RequestObjectStates into RequestConnectedObjects and RequestLocatedObjectStates
Split EvtObjectAvailable into EvtObjectConnected and EvtObjectLocated

Pose now tracks “is_accurate” property (False if pose marked Dirty by engine)
Added default constructor for Pose for an initial invalid Pose on connection
Pose copies origin_id for poses based on other poses
Pose origin_id defaults to -1 (invalid) instead of 0 if not specified.

RobotDelocalized now invalidates all object poses.
RobotDeletedLocatedObject invalidates the pose of that object.

Removed world._delete_all_objects (no longer supported in Engine, and was unused)

### DIFF
--- a/src/cozmo/conn.py
+++ b/src/cozmo/conn.py
@@ -80,11 +80,12 @@ FORCED_ROBOT_MESSAGES = {"AnimationAborted",
                          "BlockPoolDataMessage",
                          "CarryStateUpdate",
                          "ChargerEvent",
+                         "ConnectedObjectStates",
                          "CubeLightsStateTransition",
                          "CurrentCameraParams",
                          "LoadedKnownFace",
+                         "LocatedObjectStates",
                          "ObjectProjectsIntoFOV",
-                         "ObjectStates",
                          "ReactionaryBehaviorTransition",
                          "RobotChangedObservedFaceID",
                          "RobotCliffEventFinished",
@@ -372,10 +373,16 @@ class CozmoConnection(event.Dispatcher, clad_protocol.CLADProtocol):
                 'cozmoclad_version=%s app_build_version=%s',
                 version.__version__, cozmoclad.__version__, msg.buildVersion)
 
-        # We send RequestObjectStates before refreshing the animation names
-        # as this ensures that we will receive the responses before we mark the
-        # robot as ready
-        msg = _clad_to_engine_iface.RequestObjectStates()
+        # We send RequestConnectedObjects and RequestLocatedObjectStates before
+        # refreshing the animation names as this ensures that we will receive
+        # the responses before we mark the robot as ready
+        # Request information on connected objects (e.g. the object ID of each cube)
+        # (this won't provide location/pose info)
+        msg = _clad_to_engine_iface.RequestConnectedObjects()
+        self.send_msg(msg)
+        # Request the pose information for all objects whose location we know
+        # (this won't include any objects where the location is currently not known)
+        msg = _clad_to_engine_iface.RequestLocatedObjects()
         self.send_msg(msg)
 
         self.anim_names.refresh()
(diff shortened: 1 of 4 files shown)


src/cozmo/conn.py
```diff
@@ -372,10 +373,16 @@ def _handle_ui_device_connected(self, msg):
                 'cozmoclad_version=%s app_build_version=%s',
                 version.__version__, cozmoclad.__version__, msg.buildVersion)
 
-        # We send RequestObjectStates before refreshing the animation names
-        # as this ensures that we will receive the responses before we mark the
-        # robot as ready
-        msg = _clad_to_engine_iface.RequestObjectStates()
+        # We send RequestConnectedObjects and RequestLocatedObjectStates before
+        # refreshing the animation names as this ensures that we will receive
+        # the responses before we mark the robot as ready
```
msintov: Let's add a period after "ready" and maybe a blank line to separate this from the next line.

src/cozmo/world.py
```diff
@@ -345,13 +345,44 @@ def _recv_msg_object_tapped(self, evt, *, msg):
...
+            obj = self._objects.get(object_state.objectID)
+            if not obj:
+                obj = self._allocate_object_from_msg(object_state)
+            if obj:
+                obj._handle_located_object_state(object_state)
+            updated_objects.add(object_state.objectID)
+        # verify that all objects not received have invalidated poses
+        for id, obj in self._objects.items():
+            if (id not in updated_objects) and obj.pose.is_valid:
+                logger.warn("Object %s still has a valid pose but wasn't part of located_object_state" % obj)
```
gwatts: should use comma instead of percent to pass the format values `logger.warn('... %s ...', obj)`

src/cozmo/world.py
```diff
@@ -345,13 +345,44 @@ def _recv_msg_object_tapped(self, evt, *, msg):
...
+        if obj is None:
+            logger.warn("Ignoring deleted_located_object for unknown object ID %s" % msg.objectID)
+        else:
+            logger.info("Invalidating pose for deleted located object %s" % obj)
+            obj.pose.invalidate()
+
+    def _recv_msg_robot_delocalized(self, evt, *, msg):
+        # Invalidate the pose for every object
+        logger.info("Robot delocalized - invalidating poses for all objects")
+        for _, obj in self._objects.items():
```
gwatts: can use `for obj in self._object.values():` if you don't need the key/id

src/cozmo/world.py
```diff
@@ -345,13 +345,44 @@ def _recv_msg_object_tapped(self, evt, *, msg):
...
         for object_state in msg.objects:
             obj = self._objects.get(object_state.objectID)
             if not obj:
                 obj = self._allocate_object_from_msg(object_state)
             if obj:
-                obj._handle_object_state(object_state)
+                obj._handle_connected_object_state(object_state)
+
+    def _recv_msg_located_object_states(self, evt, *, msg):
+        # This is received on startup as a response to RequestLocatedObjectStates
```
msintov: Also would be good to have a period after RequestLocatedObjectStates.

src/cozmo/util.py
```diff
@@ -297,19 +297,27 @@ class Pose:
     Only poses of the same origin_id can safely be compared or operated on
     '''
 
-    __slots__ = ('_position', '_rotation', '_origin_id')
+    __slots__ = ('_position', '_rotation', '_origin_id', '_is_accurate')
 
-    def __init__(self, x, y, z, q0=None, q1=None, q2=None, q3=None, angle_z=None, origin_id=0):
+    def __init__(self, x, y, z, q0=None, q1=None, q2=None, q3=None,
+                 angle_z=None, origin_id=-1, is_accurate=True):
```
msintov: origin_id default value has changed from 0 to -1. Should we call this detail out in the release notes?

MarkWesleyAtAnki: It shouldn't affect anything user-facing, but I've updated the PR message to include this detail.

PR #90, opened 2017-03-21
COZMO-10283 Fix usage of deprecated methods from SDK

MarkWesleyAtAnki:
logger.warn is deprecated, replaced all instances with logger.warning
Flask’s send_file complains on 0.11.0 if add_etags is used without a filename, and from 0.12.0 on ignores it, so just set add_etags=False as we don’t require them and there’s no way to specify the filename _and_ a bytesIO object in a way that will use the filename for the tag.

### DIFF
--- a/src/cozmo/robot.py
+++ b/src/cozmo/robot.py
@@ -356,10 +356,10 @@ class SetLiftHeight(action.Action):
         super().__init__(**kw)
 
         if height < 0.0:
-            logger.warning("lift height %s too small, should be in 0..1 range - clamping" % height)
+            logger.warning("lift height %s too small, should be in 0..1 range - clamping", height)
             self.lift_height_mm = MIN_LIFT_HEIGHT_MM
         elif height > 1.0:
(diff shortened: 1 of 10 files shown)


src/cozmo/robot.py
```diff
@@ -356,10 +356,10 @@ def __init__(self, height, max_speed, accel, duration, **kw):
         super().__init__(**kw)
 
         if height < 0.0:
-            logger.warn("lift height %s too small, should be in 0..1 range - clamping" % height)
+            logger.warning("lift height %s too small, should be in 0..1 range - clamping" % height)
             self.lift_height_mm = MIN_LIFT_HEIGHT_MM
         elif height > 1.0:
-            logger.warn("lift height %s too large, should be in 0..1 range - clamping" % height)
+            logger.warning("lift height %s too large, should be in 0..1 range - clamping" % height)
```
gwatts: Might to replace `%` with `,` while in here to be consistent with other calls

PR #101, opened 2017-04-08
Rename ActionResult::CANCELLED to ActionResult::CANCELLED_WHILE_RUNNING

alchaussee:


MarkWesleyAtAnki:
@alchaussee Was this just a rename, or are there now multiple "Cancelled" types?

alchaussee:
ActionResult::NOT_STARTED is now considered a "cancelled" type result since the only way it can occur is if an action was cancelled/destroyed before it started running. Previously it was an abort type result which was an issue since an abort type result means that the action has completely failed and can not continue. A lot of behaviors do things on abort failures like mark blocks as "did fail to use" but if the action never started that doesn't make a lot of sense.

PR #103, opened 2017-04-13
COZMO-10139 Remove Custom objects/definitions support

MarkWesleyAtAnki:


src/cozmo/world.py
```diff
@@ -611,13 +611,64 @@ def send_available_objects(self):
...
+        for id, obj in list(self._objects.items()):
+            if isinstance(obj, objects.FixedCustomObject):
+                logger.info("Removing FixedCustomObject instance: id %s = obj '%s'", id, obj)
+                del self._objects[id]
+
     async def delete_all_custom_objects(self):
-        """Causes the robot to forget about all custom objects it currently knows about."""
+        """Causes the robot to forget about all custom (fixed + marker) objects it currently knows about.
+        
+        Note: This includes all fixed objects, and all custom marker object instances,
```
msintov: Maybe reword to "This includes all fixed custom objects"....

src/cozmo/world.py
```diff
@@ -611,13 +611,64 @@ def send_available_objects(self):
...
+
+        Note: This removes fixed custom objects only, it does NOT remove
+        the custom marker object instances or definitions.
+        """
+        msg = _clad_to_engine_iface.DeleteFixedCustomObjects()
+        self.conn.send_msg(msg)
+        await self.wait_for(_clad._MsgRobotDeletedFixedCustomObjects)
+        self._remove_fixed_custom_object_instances()
+
+    async def undefine_all_custom_marker_objects(self):
```
msintov: Do we also need an undefine for fixed_custom_objects?

MarkWesleyAtAnki: No, they're not defined - the instances are added to the world but there's no equivalent of the custom marker object definitions (which state e.g. there's a box X,Y,Z with these markers on it).

src/cozmo/world.py
```diff
@@ -611,13 +611,64 @@ def send_available_objects(self):
...
+        """Remove all custom marker object definitions, and any instances of them in the world."""
+        msg = _clad_to_engine_iface.UndefineAllCustomMarkerObjects()
+        self.conn.send_msg(msg)
+        await self.wait_for(_clad._MsgRobotDeletedCustomMarkerObjects)
+        self._remove_custom_marker_object_instances()
+        # Remove all custom object definitions / archetypes
+        self.custom_objects = {}
```
gwatts: It might be better to do `self.custom_objects.clear()` to modify the existing dict, rather than creating a new one - that way if anyone has assigned it to a local variable for easy reference, it'll continue to refer to the right thing.

PR #104, opened 2017-04-15
COZMO-8803 Replace name_face Action with Behavior

MarkWesleyAtAnki:
Face.name_face was removed in a recent update as the Action to perform this internally was removed. This has been replaced with a Behavior, so the method has been revived to use that.
To enable the use of a Behavior in a way closer to an Action, the behavior code has been updated to track changes in behavior to allow to wait for a behavior to complete, and also catch when a behavior never starts (e.g. if it is not considered runnable if various engine preconditions are not met).

### DIFF
--- a/src/cozmo/behavior.py
+++ b/src/cozmo/behavior.py
@@ -56,7 +56,7 @@ BEHAVIOR_REQUESTED = 'behavior_requested'
 #: string: Behavior running state
 BEHAVIOR_RUNNING = 'behavior_running'
 
-#: string: Behavior succeeded state
+#: string: Behavior stopped state
 BEHAVIOR_STOPPED = 'behavior_stopped'
 
 
@@ -100,7 +100,7 @@ class Behavior(event.Dispatcher):
         if self._state != BEHAVIOR_REQUESTED:
             # has not been requested (is an unrelated behavior transition)
             if self.is_running:
-                logger.warning("Behavior '%s' reported started when already running!?")
+                logger.warning("Behavior '%s' unexpectedly reported started when already running")
             return
         self._state = BEHAVIOR_RUNNING
         self.dispatch_event(EvtBehaviorStarted, behavior=self, behavior_type_name=self.type.name)
--- a/src/cozmo/behavior.py
+++ b/src/cozmo/behavior.py
@@ -33,7 +33,9 @@ behaviors.
 '''
 
 # __all__ should order by constants, event classes, other classes, functions.
-__all__ = ['EvtBehaviorStarted', 'EvtBehaviorStopped',
+__all__ = ['BEHAVIOR_IDLE', 'BEHAVIOR_REQUESTED', 'BEHAVIOR_RUNNING',
+           'BEHAVIOR_STOPPED',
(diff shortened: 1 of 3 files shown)


src/cozmo/behavior.py
```diff
@@ -45,8 +47,27 @@
...
+#: string: Behavior idle state (not requested to run)
+BEHAVIOR_IDLE = 'behavior_idle'
+
+#: string: Behavior requested state (waiting for engine to start it)
+BEHAVIOR_REQUESTED = 'behavior_requested'
+
+#: string: Behavior running state
+BEHAVIOR_RUNNING = 'behavior_running'
+
+#: string: Behavior succeeded state
```
msintov: Is it possible to request that the behavior stops after it has been started? If so, then maybe this wouldn't be the succeeded state?

src/cozmo/behavior.py
```diff
@@ -68,29 +88,85 @@ def __init__(self, robot, behavior_type, is_active=False, **kw):
...
+            self.dispatch_event(EvtBehaviorRequested, behavior=self, behavior_type_name=self.type.name)
 
     def __repr__(self):
         return '<%s type="%s">' % (self.__class__.__name__, self.type.name)
 
+    def _on_engine_started(self):
+        if self._state != BEHAVIOR_REQUESTED:
+            # has not been requested (is an unrelated behavior transition)
+            if self.is_running:
+                logger.warning("Behavior '%s' reported started when already running!?")
```
msintov: Would be good to remove '!?" from the warning and instead express that this is "unexpected" or something.

src/cozmo/faces.py
```diff
@@ -359,21 +357,30 @@ def _recv_msg_robot_erased_enrolled_face(self, evt, *, msg):
...
-    #         name (string): The name that will be assigned to this face
-    #     Returns:
-    #         An instance of :class:`cozmo.faces.EnrollNamedFace` action object
-    #     '''
-    #     logger.info("Sending enroll named face request for face=%s and name=%s", self, name)
-    #     action = self.enroll_named_face_factory(face=self, name=name, conn=self.conn,
-    #                                             robot=self._robot, dispatch_parent=self)
-    #     self._robot._action_dispatcher._send_single_action(action)
-    #     return action
+    def name_face(self, name):
```
msintov: Do we want to disallow any characters for name?

src/cozmo/robot.py
```diff
@@ -1314,11 +1332,15 @@ def start_behavior(self, behavior_type):
...
+
+        if self._current_behavior is not None:
+            self._current_behavior._set_stopped()
+
         b = self.behavior_factory(self, behavior_type, is_active=True, dispatch_parent=self)
         msg = _clad_to_engine_iface.ExecuteBehaviorByExecutableType(
                 behaviorType=behavior_type.id)
         self.conn.send_msg(msg)
-        self._is_behavior_running = True
+        self._current_behavior = b
```
msintov: I know this is pre-existing but would be good to rename the var from 'b' to 'behavior' or something more explanatory.

PR #116, opened 2017-05-23
COZMO-11655 Update IFTTT aiohttp scripts to work with latest aiohttp version

msintov:
Tested with aiohttp versions 1.0.5 and 2.0.7

### DIFF
--- a/examples/if_this_then_that/ifttt_gmail.py
+++ b/examples/if_this_then_that/ifttt_gmail.py
@@ -54,7 +54,7 @@ Follow these steps to set up and run the example:
             3. Under "Choose a Trigger", select “Any new email in inbox".
         d) Set up your action.
             1. Click “that".
-            2. Select “Maker" to set it as your action channel. Connect to the Maker channel if prompted.
+            2. Select “Maker Webhooks" to set it as your action channel. Connect to the Maker channel if prompted.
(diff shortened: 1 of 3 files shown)


examples/if_this_then_that/ifttt_gmail.py
```diff
@@ -92,7 +92,7 @@
 from common import IFTTTRobot
 
 
-app = web.Application()
+app = web.Application(loop=asyncio.get_event_loop())
```
MarkWesleyAtAnki: passing loop into Application looked like it’s deprecated in latest aiohttp, so while this works for now (with warning i think) it might not forever. It looks like `run_app` will default to the loop, but that happens after we try to access app.loop - so the better fix might be to revert that, and later where we have:
```python
         sdk_conn = cozmo.connect_on_loop(app.loop)
         # Wait for the robot to become available and add it to the app object.
         app['robot'] = app.loop.run_until_complete(sdk_conn.wait_for_robot())
```

instead do:

```python
         app_loop = asyncio.get_event_loop()  # app.loop isn't set until run_app on latest versions of aiohttp
         sdk_conn = cozmo.connect_on_loop(app_loop)
         # Wait for the robot to become available and add it to the app object.
         app['robot'] = app_loop.run_until_complete(sdk_conn.wait_for_robot())
```

PR #118, opened 2017-05-30
COZMO-11731 Actions handling OffTreadsState on their own

mmichini:
Added new Action Result “INVALID_OFF_TREADS_STATE” for situations where the robot ends up in an invalid orientation while executing an action.


mmichini:
should go to nextclad, not master.

PR #121, opened 2017-05-31
Nicolas/cozmo 8481 cozmo 11263 exposed new actions with examples

ankiNicolas:


src/cozmo/alignment.py
```diff
@@ -0,0 +1,55 @@
...
+
+        LiftFinger = _AlignmentType("LiftFinger", _clad_to_engine_cozmo.AlignmentType.LIFT_FINGER)
+
+        LiftPlate = _AlignmentType("LiftPlate", _clad_to_engine_cozmo.AlignmentType.LIFT_PLATE)
+
+        Body = _AlignmentType("Body", _clad_to_engine_cozmo.AlignmentType.BODY)
+
+        Custom = _AlignmentType("Custom", _clad_to_engine_cozmo.AlignmentType.CUSTOM)
+
+        _id_to_behavior_type = dict()
```
MarkWesleyAtAnki: should be _id_to_robot_alignment_type. (Basically anywhere you have Behavior in this file it should be renamed to e.g. RobotAlignment, as those names were from the BehaviorType equivalent of this).

src/cozmo/alignment.py
```diff
@@ -0,0 +1,55 @@
...
+
+        @classmethod
+        def find_by_id(cls, id):
+            return cls._id_to_behavior_type.get(id)
+
+    # populate AlignmentTypes _id_to_alignment_type mapping
+    for (_name, _bt) in AlignmentTypes.__dict__.items():
```
MarkWesleyAtAnki: the _bt was shorthand for BehaviorType, so again should be _at or _rat for robot-alignment-type

src/cozmo/alignment.py
```diff
@@ -0,0 +1,55 @@
...
+        if isinstance(_bt, _AlignmentType):
+            AlignmentTypes._id_to_behavior_type[_bt.id] = _bt
+
+except AttributeError as exc:
+    err = ('Incorrect version of cozmoclad package installed.  '
```
MarkWesleyAtAnki: This is good for now, but longer term I think we should throw an error on import of cozmo if the cozmoclad version is incorrect (the try/except blocks here were mostly to catch if cozmoclad was so incompatible that "_clad_to_engine_cozmo.AlignmentType.LIFT_PLATE" didn't exist.

src/cozmo/robot.py
```diff
@@ -36,7 +36,8 @@
 # __all__ should order by constants, event classes, other classes, functions.
 __all__ = ['MAX_HEAD_ANGLE', 'MIN_HEAD_ANGLE', 'MIN_LIFT_HEIGHT_MM', 'MAX_LIFT_HEIGHT_MM',
            'EvtRobotReady',
-           'GoToPose', 'DisplayOledFaceImage', 'DriveOffChargerContacts',
+           'GoToPose', 'GoToObject', 'DockWithCube', 'RollCube',
```
MarkWesleyAtAnki: These should be in alphabetical order (makes it easier to spot if something is missing)

src/cozmo/robot.py
```diff
@@ -130,6 +131,55 @@ def _encode(self):
...
+        self.obj = obj
+        #: bool: whether to check if there is an object on top
+        self.check_for_object_on_top = check_for_object_on_top
+
+    def _repr_values(self):
+        return "object=%s, check_for_object_on_top=%s" % (self.obj, self.check_for_object_on_top)
+
+    def _encode(self):
+        return _clad_to_engine_iface.RollObject(objectID=self.obj.object_id,
+                                                approachAngle_rad = 0, #I've chosen not to expose the approach angle, as it doesn't seem to have any effect
```
ankiNicolas: I spent some time trying to mess with approach angle with the flag set to true and couldn't figure out what effect it had.  My assumption was that it would make cozmo drive around to another side of the block, and roll it that way.  Changing the comment, though.

src/cozmo/robot.py
```diff
@@ -174,7 +174,7 @@ def _repr_values(self):
 
     def _encode(self):
         return _clad_to_engine_iface.RollObject(objectID=self.obj.object_id,
-                                                approachAngle_rad = 0, #I've chosen not to expose the approach angle, as it doesn't seem to have any effect
+                                                approachAngle_rad = 0, #Approach angle is not used, as it doesn't seem to have any effect
```
ankiNicolas: Ah, figured it out - it looks like i need to set usePreDockPose=True in order for useApproachAngle=True to work.  I'll rework this to expose more configurability.

src/cozmo/robot.py
```diff
@@ -130,6 +130,67 @@ def _encode(self):
...
+    _action_type = _clad_to_engine_cozmo.RobotActionType.ROLL_OBJECT_LOW
+
+    def __init__(self, obj, approach_angle, check_for_object_on_top, **kw):
+        super().__init__(**kw)
+        #: The object (e.g. an instance of :class:`cozmo.objects.LightCube`) that is being put down
+        self.obj = obj
+        #: bool: whether to check if there is an object on top
+        self.check_for_object_on_top = check_for_object_on_top
+        if approach_angle is None:
+            self.useApproachAngle = False
```
MarkWesleyAtAnki: python variables should be snake-cased, so should be `self.use_approach_angle = False`

src/cozmo/robot.py
```diff
@@ -130,6 +130,67 @@ def _encode(self):
...
+            self.approachAngle = util.degrees(0)
```
MarkWesleyAtAnki: should be self.approach_angle

src/cozmo/robot.py
```diff
@@ -130,6 +130,67 @@ def _encode(self):
...
+        else:
+            self.useApproachAngle = True
+            self.approachAngle = approach_angle
+
+    def _repr_values(self):
+        return "object=%s, check_for_object_on_top=%s" % (self.obj, self.check_for_object_on_top)
```
MarkWesleyAtAnki: Minor, but could add approach angle info here too

src/cozmo/robot.py
```diff
@@ -130,6 +130,67 @@ def _encode(self):
...
+    Returned by :meth:`~cozmo.robot.Robot.dock_with_cube`
+    '''
+    _action_type = _clad_to_engine_cozmo.RobotActionType.ALIGN_WITH_OBJECT
+
+    def __init__(self, obj, approach_angle, **kw):
+        super().__init__(**kw)
+        #: The object (e.g. an instance of :class:`cozmo.objects.LightCube`) that is being put down
+        self.obj = obj
+        if approach_angle is None:
+            self.useApproachAngle = False
```
MarkWesleyAtAnki: python variables should be snake-cased, so should be self.use_approach_angle = False
and self.approach_angle =

src/cozmo/robot.py
```diff
@@ -130,6 +130,67 @@ def _encode(self):
...
+            self.approachAngle = util.degrees(0)
+        else:
+            self.useApproachAngle = True
+            self.approachAngle = approach_angle
+
+    def _repr_values(self):
+        return "object=%s" % (self.obj)
```
MarkWesleyAtAnki: minor, but could add approach angle to this too

src/cozmo/robotAlignment.py
```diff
@@ -0,0 +1,54 @@
+# Copyright (c) 2017 Anki, Inc.
```
MarkWesleyAtAnki: file should be renamed to be snake-cased - so robot_alignment.py

examples/tutorials/04_cubes_and_objects/11_dock_with_cube.py
```diff
@@ -14,25 +14,23 @@
...
+picking it up.  You must place a cube in front of Cozmo so that he can see it.
 '''
 
-import asyncio
 import cozmo
 
-async def dock_with_cube_test(robot: cozmo.robot.Robot):
-    '''The core of the dock with cube test program'''
+async def dock_with_cube(robot: cozmo.robot.Robot):
+
```
MarkWesleyAtAnki: remove empty line at top of method

examples/tutorials/04_cubes_and_objects/12_roll_cube.py
```diff
@@ -14,28 +14,22 @@
...
+You must place a cube in front of Cozmo so that he can see it.
 '''
 
-import asyncio
 import cozmo
 
-async def roll_cube_test(robot: cozmo.robot.Robot):
-    '''The core of the dock with cube test program'''
+async def roll_a_cube(robot: cozmo.robot.Robot):
+
```
MarkWesleyAtAnki: remove empty line at top of method

src/cozmo/robot.py
```diff
@@ -1666,14 +1662,15 @@ def go_to_object(self, target_object, distance_from_object,
...
+                       distance_from_marker=None,
                        in_parallel=False, num_retries=0):
         '''Tells Cozmo to dock with a specified cube object.
 
         Args:
-            target_object (:class:`cozmo.objects.LightCube`): The destination object.
-            approach_angle (:class:`cozmo.util.Angle`): The angle to approach the cube from.
-            alignment_type (:class:`cozmo.robot_alignment.RobotAlignmentTypes`): which part of the robot to line up with the object
+            target_object (:class:`cozmo.objects.LightCube`): The cube to dock with.
+            approach_angle (:class:`cozmo.util.Angle`): The angle to approach the cube from.  For example, 180 will cause cozmo to drive past the cube and approach it from behind.
```
MarkWesleyAtAnki: "180" is ambiguous here as the angle type doesn't define the units (could be degrees or radians depending on how they called it), so suggest changing the doc here to say "180 degrees"

src/cozmo/robot.py
```diff
@@ -1700,9 +1697,9 @@ def roll_cube(self, target_object, approach_angle=None, check_for_object_on_top=
         '''Tells Cozmo to roll a specified cube object.
 
         Args:
-            target_object (:class:`cozmo.objects.LightCube`): The destination object.
-            approach_angle (:class:`cozmo.util.Angle`): The angle to approach the cube from.
-            check_for_object_on_top (bool): only roll the cube if there isn't a cube on top of it
+            target_object (:class:`cozmo.objects.LightCube`): The cube to roll.
+            approach_angle (:class:`cozmo.util.Angle`): The angle to approach the cube from.   For example, 180 will cause cozmo to drive past the cube and approach it from behind.
```
MarkWesleyAtAnki: "180 degrees"

src/cozmo/robot.py
```diff
@@ -1700,9 +1697,9 @@ def roll_cube(self, target_object, approach_angle=None, check_for_object_on_top=
...
+            check_for_object_on_top (bool): only roll the cube if there isn't a cube on top of it.  If there is the cube will be ignored.
```
MarkWesleyAtAnki: I don't think "If there is the cube will be ignored."  is correct, suggest changing to e.g. "If there is a cube on top, and check_for_object_on_top is True, then Cozmo will ignore the action"

PR #129, opened 2017-06-14
COZMO-12298 clean up sdk for lint

ankiNicolas:
After reducing down the pylint pass to avoid spamming style-type errors, there were a few remaining things that I wanted either to fix or locally suppress

### DIFF
--- a/src/cozmo/world.py
+++ b/src/cozmo/world.py
@@ -672,7 +672,7 @@ class World(event.Dispatcher):
         """
         msg = _clad_to_engine_iface.DeleteAllCustomObjects()
         self.conn.send_msg(msg)
-        # suppression for _MsgRobotDeletedAllCustomObjects "no-memeber" on pylint
+        # suppression for _MsgRobotDeletedAllCustomObjects "no-member" on pylint
         #pylint: disable=no-member
         await self.wait_for(_clad._MsgRobotDeletedAllCustomObjects)
         self._remove_custom_marker_object_instances()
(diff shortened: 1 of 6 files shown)


src/cozmo/world.py
```diff
@@ -979,6 +985,7 @@ def disconnect_from_cubes(self):
         Returns:
             bool: True if all 3 cubes are now connected.
         """
+        #pylint: disable=no-member
```
MarkWesleyAtAnki: why is this needed here - world is fairly straightforward i think

src/cozmo/world.py
```diff
@@ -926,6 +931,7 @@ def _remove_fixed_custom_object_instances(self):
         Returns:
             A :class:`cozmo.objects.FixedCustomObject` instance with the specified dimensions and pose.
         '''
+        #pylint: disable=no-member
```
MarkWesleyAtAnki: does pylint require that these are at the top of the file, or can they be placed next to the problem line - would make it easier to see what line is causing the issue.

src/cozmo/world.py
```diff
@@ -672,6 +672,8 @@ def _remove_fixed_custom_object_instances(self):
         """
         msg = _clad_to_engine_iface.DeleteAllCustomObjects()
         self.conn.send_msg(msg)
+        # suppression for _MsgRobotDeletedAllCustomObjects "no-memeber" on pylint
```
MarkWesleyAtAnki: typo "memeber" - maybe add "(the class is created at startup in _clad)"

PR #131, opened 2017-06-16
COZMO-12173 Quizmaster Cozmo example

MarkWesleyAtAnki:
Add an example quiz program using the cubes as buzzers
Add Json files to the examples zip (as quiz uses one)
Updated run.py to more cleanly handle exit of async programs when the user presses CTRL-C to keyboard interrupt
Added objects.LightCubeIDs helper list of all 3 cube IDs

### DIFF
--- a/examples/apps/quizmaster_cozmo.py
+++ b/examples/apps/quizmaster_cozmo.py
@@ -31,38 +31,39 @@ class QuizQuestion:
     
     Args:
         question (str): The question.
-        options (list of str): 4 multiple choice answers where the
-            1st element is the correct answer. (They will be shuffled each time)
+        anwer_options (list of str): 4 multiple choice answers where the
+            1st element is the correct answer. (Choices will be shuffled each time.)
     Raises:
-        :class:`ValueError` if not supplied exactly 4 options.
+        :class:`ValueError` if not supplied exactly 4 anwer_options.
     """
-    def __init__(self, question, options):
-        if len(options) != 4:
-            raise ValueError("Expected 4 options, got %s" % len(options))
+    def __init__(self, question, anwer_options):
+        if len(anwer_options) != 4:
+            raise ValueError("Expected 4 anwer_options, got %s" % len(anwer_options))
         self.question = question
         self._answer_index = 0
-        self.options = list(options)  # copy the options, so we can shuffle them
-        self.shuffle_options()
+        self.anwer_options = list(anwer_options)  # copy the anwer_options, so we can shuffle them
+        self.shuffle_anwer_options()
 
     @property
     def answer_number(self):
-        """int: The number (1..4) representing the correct answer."""
+        """int: The number (i.e. 1, 2, 3 or 4) representing the correct answer."""
         return self._answer_index + 1
 
     @property
     def answer_str(self):
         """str: The string representing the correct answer."""
-        return self.options[self._answer_index]
+        return self.anwer_options[self._answer_index]
 
-    def shuffle_options(self):
-        """Shuffle the options so that they're not always read in the same order."""
+    def shuffle_anwer_options(self):
+        """Shuffle the anwer_options so that they're not always read in the same order."""
+
         # to shuffle whilst keeping track of the answer, we first pop the
         # answer out, shuffle the rest, and then insert the answer at a random
         # known point.
-        answer = self.options.pop(self._answer_index)
-        shuffle(self.options)
-        self._answer_index = randrange(len(self.options)+1)
-        self.options.insert(self._answer_index, answer)
+        answer = self.anwer_options.pop(self._answer_index)
+        shuffle(self.anwer_options)
+        self._answer_index = randrange(len(self.anwer_options)+1)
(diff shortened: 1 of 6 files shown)


examples/apps/quizmaster_cozmo.py
```diff
@@ -0,0 +1,403 @@
...
+        if len(options) != 4:
+            raise ValueError("Expected 4 options, got %s" % len(options))
+        self.question = question
+        self._answer_index = 0
+        self.options = list(options)  # copy the options, so we can shuffle them
+        self.shuffle_options()
+
+    @property
+    def answer_number(self):
+        """int: The number (1..4) representing the correct answer."""
```
msintov: To describe the list in a way all people should understand, could do: (i.e., 1, 2, 3 or 4). I know in the SDK we already describe lists like (1..4), but for a higher level user who might want to use this example, instead I think we should use a more typical description.

examples/apps/quizmaster_cozmo.py
```diff
@@ -0,0 +1,403 @@
...
+        return self._answer_index + 1
+
+    @property
+    def answer_str(self):
+        """str: The string representing the correct answer."""
+        return self.options[self._answer_index]
+
+    def shuffle_options(self):
```
msintov: Again, could rename to shuffle_answer_options or similar.

examples/apps/quizmaster_cozmo.py
```diff
@@ -0,0 +1,403 @@
...
+        """Shuffle the options so that they're not always read in the same order."""
+        # to shuffle whilst keeping track of the answer, we first pop the
```
msintov: Could add a line of whitespace above the # comment line so it's more readable.

examples/apps/quizmaster_cozmo.py
```diff
@@ -0,0 +1,403 @@
...
+        self.options.insert(self._answer_index, answer)
+
+
+class CozmoQuizPlayer:
+    """A player in the quiz.
+
+    Args:
+        robot (:class:`cozmo.robot.Robot`): The cozmo robot.
+        cube (:class:`cozmo.objects.LightCube`): This player's cube.
+        index(int): The number (0..2) specifying the index of this player and cube.
```
msintov: Again, could write the list instead as "(i.e., 0, 1, or 2)"

examples/apps/quizmaster_cozmo.py
```diff
@@ -0,0 +1,403 @@
...
+        self.name = name
+        self.score = 0
+        self._has_buzzed_in = False
+        self._answer_index = 0
+
+    def verify_setup(self):
+        # Return True if and only if the player was setup correctly and has a connected cube.
+        success = True
+        if self._cube is None:
+            cozmo.logger.warning("Cozmo is not connected to a cube %s - check the battery.", (self._index+1))
```
msintov: Is a battery failure the most common reason the value would be None?

examples/apps/quizmaster_cozmo.py
```diff
@@ -0,0 +1,403 @@
...
+                conjunction = " or " if is_last_option else ", "
+            text += conjunction + str(i+1) + ": " + list_of_options[i]
+        return text
+
+    def say_text(self, text, in_parallel=False):
+        print("%s" % text)
+        return self._robot.say_text(text, in_parallel=in_parallel)
+
+    async def wait_for_answer(self, player):
+        # Wait for player's answer (whatever they leave selected after x seconds)
```
msintov: "they" is confusing here. Could say "whatever the player leaves selected..."

examples/apps/quizmaster_cozmo.py
```diff
@@ -0,0 +1,403 @@
...
+        winning_players = []
+        for player in self._players:
+            if len(winning_players) == 0 or player.score > winning_players[0].score:
+                winning_players = [player]
+            elif player.score == winning_players[0].score:
+                winning_players.append(player)
+        return winning_players
+
+    async def report_leader(self, is_final_score):
+        # report however is currently winning / won
```
msintov: Could rewrite as "Report the winning player" or similar. "however" should be "whoever" if original text is kept.

examples/apps/quizmaster_cozmo.py
```diff
@@ -0,0 +1,403 @@
...
+            else:
+                # Incorrect
+                player.score -= 1
+                action = self._robot.play_anim_trigger(cozmo.anim.Triggers.KnockOverFailure)
+                self._answering_player = None
+                player.turn_light_off()
+                await action.wait_for_completed()
+
+    async def ask_question(self, question: QuizQuestion):
+        # Reset everyone for the question
```
msintov: Comment might be clearer as "Reset for a new question"

examples/apps/quizmaster_cozmo.py
MarkWesleyAtAnki: i think so - generally it's because either the cube isn't there (e.g. they lost it) or it's battery is dead

PR #138, opened 2017-07-01
COZMO-12820 doc update

ankiNicolas:
vagrant instructions were missing that "vagrant ssh" should be run after "vagrant up"

### DIFF
--- a/docs/source/vagrant.rst
+++ b/docs/source/vagrant.rst
@@ -45,7 +45,6 @@ Install
     2. Enter the following commands::
 
         vagrant up
-        vagrant ssh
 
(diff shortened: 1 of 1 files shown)


docs/source/vagrant.rst
```diff
@@ -74,6 +74,8 @@ Install
 
       cd /vagrant/cozmo_sdk_examples_***
 
+    .. important:: You must either relaunch with "vagrant up" or save your virtual machine's state when shutting down.  Otherwise the /vagrant/ folder on the virtual machine will be empty on subsequent runs.
```
MarkWesleyAtAnki: Instead of this information being duplicated in both Windows and Mac steps 5/6, might be better to just have this as a note e.g. after step 9

PR #142, opened 2017-08-09
COZMO-10308 Added Navigation Memory Map access

MarkWesleyAtAnki:
Added nav_memory_map.py - supports deserializing a Navigation Memory Map from the engine, and providing access to query the contents at a given X,Y position.

Lots of other tidy up including:

Importing cozmo now sys.exits if the Python version is too old, or if the cozmoclad version is older than the minimum version supported by the cozmo code - this is to simplify error handling in other code that relied on specific Clad message names being present or in a given format.
Zero-length CLAD messages can now be received
Renamed Rotation to Quaternion in util (but kept Rotation as an alias to avoid breaking any external code).
Added a Matrix44 class, with conversion functions for converting Pose and Quaternion objects into a Matrix format (which is generally easier to use in 3D graphics and for general 3D math)
Factored out common CLAD enum handling into a new CladEnumWrapper helper class - updated BehaviorTypes to use this, and it's used in the new nav_memory_map's NodeContentTypes
Object annotation in the 2D camera view now supports per-class specialization, so that cubes and custom objects can add additional useful info like CubeId and ObjectType.
Face annotation now includes the expression score (if facial expression scores have been requested), UNKNOWN expressions are now handled correctly (they were showing up as zero scored Neutral faces before)
Added type info to the rest of the robot.Robot class to help auto-completion in IDEs.
Added support for optionally ignoring any combination of head, lift and body (wheels/treads) animation tracks when playing an animation trigger. Updated the 08_animation.py example to include an example.
Added support for "use_lift_safe" on playing an animation that automatically allows the lift track to be ignored if Cozmo is currently carrying an object.
Added mobile device IMU support, and the remote_control_cozmo example requests the mobile device's IMU (accelerometer and gyro) data and displays it in the display. (This data could be used to e.g. control Cozmo via the mobile device).
Disabled logging on action aborting by default (can be enabled on a per-call basis by passing in log_abort_messages=True) as it spams a lot when heavily used.
Chmod +x quick_tap.py and color_finder.py so that they can be run directly from command line
Fix bug in quick_tap.py if you tapped a cube before Cozmo chose his cube.


examples/apps/remote_control_cozmo.py
```diff
@@ -71,6 +71,18 @@ def print_line(text_line):
         print_line('Accelmtr: <%.1f, %.1f, %.1f>' % robot.accelerometer.x_y_z)
         print_line('Gyro: <%.1f, %.1f, %.1f>' % robot.gyro.x_y_z)
 
+        # Display the Accelerometer and Gyro data for the mobile device
+
+        if robot.device_accel_raw is not None:
+            print_line('Device Acc Raw: <%.2f, %.2f, %.2f>' % robot.device_accel_raw.x_y_z)
```
msintov: These print_line lines aren't too noisy in the output?

src/cozmo/annotate.py
```diff
@@ -307,7 +307,7 @@ def label_for_face(self, obj):
         '''
         expression = obj.known_expression
         if len(expression) > 0:
-            expression += " "
+            expression += "=%s%% " % obj.expression_score
```
msintov: What does this line do?

src/cozmo/nav_memory_map.py
```diff
@@ -0,0 +1,316 @@
...
+        #: list of :class:`NavMemoryMapGridNode`: ``None`` for leaf nodes, a list of 4
+        #: child nodes otherwise.
+        self.children = None
+
+        #: An attribute of :class:`NodeContentTypes`: The content type in this
+        #: node. Only leaf nodes have content, this is ``None`` for all other
+        #: nodes.
+        self.content = None  # type: _NodeContentType
+
+        self._next_child = 0  # Used when building to track which branch to follow (
```
msintov: Looks like an unwanted trailing ( here

src/cozmo/nav_memory_map.py
```diff
@@ -0,0 +1,316 @@
...
+        dist_y = abs(self.center.y - y)
+        return (dist_x <= half_size) and (dist_y <= half_size)
+
+    def get_node(self, x, y, _assumed_in_bounds=False):
+        """Get the node at the given x,y coordinates.
+
+        Args:
+            x (float): x coordinate for the point
+            y (float): y coordinate for the point
+
```
msintov: Could add Arg _assumed_in_bounds

src/cozmo/nav_memory_map.py
```diff
@@ -0,0 +1,316 @@
...
+    def _add_child(self, content, depth):
+        """Add a child node to the quad tree.
+
+        The quad-tree is serialized to a flat list of nodes, we deserialize
+        back to a quad-tree structure here, with the depth of each node
+        indicating where it is placed.
+
+        Args:
+            content (:class:`_NodeContentType`):
+            depth (int):
```
msintov: Missing depth description

src/cozmo/nav_memory_map.py
```diff
@@ -0,0 +1,316 @@
...
+
+        Returns:
+            bool: True if parent should use the next child for future _add_child
+            calls (this is an internal implementation detail of h
```
msintov: Whoops, this return parenthesis is cut off

examples/apps/remote_control_cozmo.py
MarkWesleyAtAnki: It's "printing" to the camera image, not the output.

src/cozmo/annotate.py
MarkWesleyAtAnki: if there is a specific known expression, then also show the score (by appending it to the string, along with a % symbol to show that the value is out of 100 (`%%` is the escaped value for a `%` vs `%d` `%s` etc.)

src/cozmo/nav_memory_map.py
MarkWesleyAtAnki: I didn't want to expose that as it's an implementation detail, but I'll rework it to use a private method to hide it a bit better.

src/cozmo/nav_memory_map.py
MarkWesleyAtAnki: Good catch - I think I was halfway through writing the docstring for this when I decided I should just rename the function to be private as it's entirely an internal implementation detail, and then because I reviewed the docs based on the generated docs (and private methods automatically don't publish docstrings) I didn't notice that it was still a mess.

PR #143, opened 2017-08-16
color-finder update: uncommented lookaround state, added "swatting"

maxcembalest:
"swatting" feature brings the lift down when tennis ball is close to Cozmo (really when the color_to_find takes up a large portion of the screen), to simulate Cozmo swatting at the thing

### DIFF
--- a/examples/apps/color_finder.py
+++ b/examples/apps/color_finder.py
@@ -650,7 +650,7 @@ class BlobDetector():
             longest_points_list = functools.reduce(lambda largest, current: largest if (largest > current) else current, values)
             sample_x, sample_y = longest_points_list[0]
             largest_blob_key = self.keys.at(sample_x, sample_y).value
-            self.update_largest_blob_size(largest_blob_key)
+            self.largest_blob_size = len(self.blobs_dict[largest_blob_key])
         return largest_blob_key
 
     def get_blob_center(self):
@@ -673,15 +673,11 @@ class BlobDetector():
             blob_center = (int(average_x), int(average_y))
         return blob_center
 
-    def update_largest_blob_size(self, key):
-        '''Updates self.largest_blob_size based on the current largest blob.'''
-        self.largest_blob_size = len(self.blobs_dict[key])
-
     def get_blob_size(self):
         '''Gets the number of pixels in the largest blob.
 
         Returns:
-            int specifying the size of the blob
+            int: The size, in pixels, of the largest blob
         '''
         return self.largest_blob_size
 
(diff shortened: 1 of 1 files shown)


examples/apps/color_finder.py
```diff
@@ -348,13 +349,15 @@ def get_low_res_view(self):
         downsized_image = image.resize((DOWNSIZE_WIDTH, DOWNSIZE_HEIGHT), resample = Image.LANCZOS)
         return downsized_image
 
-    def on_finding_a_blob(self, blob_center):
+    def on_finding_a_blob(self, blob_center, blob_size):
```
MarkWesleyAtAnki: please add blob_size to arg strings below

examples/apps/color_finder.py
```diff
@@ -667,6 +672,12 @@ def get_blob_center(self):
             blob_center = (int(average_x), int(average_y))
         return blob_center
 
+    def update_largest_blob_size(self, key):
```
MarkWesleyAtAnki: please add docstring

examples/apps/color_finder.py
```diff
@@ -667,6 +672,12 @@ def get_blob_center(self):
...
+        self.largest_blob_size = len(self.blobs_dict[key])
+
+    def get_blob_size(self):
```
MarkWesleyAtAnki: please add docstring

examples/apps/color_finder.py
```diff
@@ -673,9 +674,15 @@ def get_blob_center(self):
...
 
     def update_largest_blob_size(self, key):
+        '''Updates self.largest_blob_size based on the current largest blob.'''
         self.largest_blob_size = len(self.blobs_dict[key])
 
     def get_blob_size(self):
+        '''Gets the number of pixels in the largest blob.
+
+        Returns:
+            int specifying the size of the blob
```
MarkWesleyAtAnki: docstring for return value should be: "type: description" - e.g:
`int: The size, in pixels, of the largest blob.`

examples/apps/color_finder.py
```diff
@@ -673,9 +674,15 @@ def get_blob_center(self):
         return blob_center
 
     def update_largest_blob_size(self, key):
+        '''Updates self.largest_blob_size based on the current largest blob.'''
```
MarkWesleyAtAnki: that isn't strictly true - the method will set largest_blob_size to the size of the key blob. In this case, it might be best to just remove the method (as it's just 1 line), and change the calling code to:
```python
self.largest_blob_size = len(self.blobs_dict[largest_blob_key])
```

PR #150, opened 2017-09-20
Minor fixes for 2.0.1 app release

MarkWesleyAtAnki:


src/cozmo/opengl.py
```diff
@@ -569,7 +569,10 @@ def __init__(self, robot):
 
         self.face_frames = []
         for face in world._faces.values():
-            self.face_frames.append(FaceRenderFrame(face))
+            # Ignore face's that have a newer version (with updated id)
```
msintov: "face's" should be "faces"

PR #151, opened 2017-10-05
Catch import error if importing old Python 2 version of PIL

nishkar:
(need to catch `SyntaxError` in addition to `ImportError`)


nishkar:
Closing this PR to make the change in a different fork of this repo

PR #154, opened 2017-11-13
Surface SDK version errors more cleanly

MarkWesleyAtAnki:
SDKVersionMismatch can now be caught by client code, if they wish to handle it separately.
SDKVersionMismatch exception now includes SDK version along with required and actual App versions.
After detecting the version mismatch, any remaining messages on the connection are ignored (to avoid throwing exceptions when they fail to unpack)
If no valid device connection is found, and an SDKVersionMismatch was detected on any device, then that SDKVersionMismatch exception is returned instead of a general "no devices found" exception.

### DIFF
--- a/src/cozmo/run.py
+++ b/src/cozmo/run.py
@@ -252,6 +252,7 @@ class AndroidConnector(DeviceConnector):
         self._exec('-s', serial, 'forward', '--remove', self.portspec)
 
     async def connect(self, loop, protocol_factory, conn_check):
+        version_mismatch = None
         for serial in self._devices():
             if serial in self._connected:
                 continue
@@ -276,6 +277,8 @@ class AndroidConnector(DeviceConnector):
                     try:
                         await conn_check(proto)
                     except Exception as e:
+                        if isinstance(e, exceptions.SDKVersionMismatch):
+                            version_mismatch = e
                         logger.debug('Failed connection check: %s', e)
                         raise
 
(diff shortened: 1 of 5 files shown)


src/cozmo/run.py
```diff
@@ -276,6 +277,8 @@ def _remove_forward(self, serial):
                     try:
                         await conn_check(proto)
                     except Exception as e:
+                        if isinstance(e, exceptions.SDKVersionMismatch):
```
shawnblakesley: Might be better to have a separate `except exceptions.SDKVersionMismatch as e:` branch than to use `isinstance` here

src/cozmo/run.py
```diff
@@ -286,6 +289,10 @@ def _remove_forward(self, serial):
             except:
```
shawnblakesley: Using an `except` branch here for `SDKVersionMismatch` which then `raise`s that error might be cleaner than using the `version_mismatch` variable

src/cozmo/run.py
MarkWesleyAtAnki: I did originally do it that way, but I felt it was duplicating more code that way. I'm not hugely torn either way - if there's a best-of-both-worlds I'd be interested?

src/cozmo/run.py
MarkWesleyAtAnki: I want it to try each Android in turn, and only raise SDKVersionMismatch if they all fail

PR #169, opened 2017-12-06
add a throw_error flag to run_program to throw error rather than exit

AdrieanKhisbe:
I started building a REPL starting from the cli.py program in the sdk examples.

I have the need to catch connexion error to display some message in a more frendly way than logs.
Here is a patch to enable this =)

### DIFF
--- a/src/cozmo/run.py
+++ b/src/cozmo/run.py
@@ -781,7 +781,7 @@ def run_program(f, use_viewer=False, conn_factory=conn.CozmoConnection,
                 connector=None, force_viewer_on_top=False,
                 deprecated_filter="default", use_3d_viewer=False,
                 show_viewer_controls=True,
-                throw_error=False):
+                exit_on_connection_error=True):
     '''Connect to Cozmo and run the provided program/function f.
 
     Args:
@@ -810,6 +810,8 @@ def run_program(f, use_viewer=False, conn_factory=conn.CozmoConnection,
             `use_viewer` are set then the 2D camera view will render in an OpenGL
             window instead of a TkView window.
         show_viewer_controls (bool): Specifies whether to draw controls on the view.
+        exit_on_connection_error (bool): Specify whether the program should exit on
+            connection error or should an error be raised. Default to true.
     '''
     setup_basic_logging(deprecated_filter=deprecated_filter)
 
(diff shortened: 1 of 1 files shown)


src/cozmo/run.py
```diff
@@ -780,7 +780,8 @@ def setup_basic_logging(general_log_level=None, protocol_log_level=None,
 def run_program(f, use_viewer=False, conn_factory=conn.CozmoConnection,
                 connector=None, force_viewer_on_top=False,
                 deprecated_filter="default", use_3d_viewer=False,
-                show_viewer_controls=True):
+                show_viewer_controls=True,
+                throw_error=False):
```
MarkWesleyAtAnki: New arguments need documenting in the docstring just below this too (see the pattern for use_3d_viewer, show_viewer_controls etc.).
throw_error is a bit unclear - I think it would be clearer to name it "exit_on_connection_error" (defaulted to True), and explain in the docstring that if this is set to False then it will re-raise the exception instead.

MarkWesleyAtAnki:
@AdrieanKhisbe Thanks - please complete a Contributor License Agreement - see https://github.com/anki/cozmo-python-sdk/blob/master/CONTRIBUTING.md - we're unable to accept code until we've received these. (Note: You only need to complete this step once, you don't need to complete it for every Pull Request)

src/cozmo/run.py
AdrieanKhisbe: okey dokey. I see your point
Will do!

src/cozmo/run.py
```diff
@@ -847,4 +848,8 @@ def wrapper(sdk_conn):
     except KeyboardInterrupt:
         logger.info('Exit requested by user')
     except exceptions.ConnectionError as e:
-        sys.exit("A connection error occurred: %s" % e)
+        if throw_error:
+            logger.error("A connection error occurred: %s" % e)
+            raise e
```
MarkWesleyAtAnki: Also note, as per our coding standard:
"To re-raise the original exception use raise by itself or raise MyException() from exc (rather than raise exc) to maintain the original stack trace."
(so this should probably just be `raise`)

AdrieanKhisbe: Okey Dokey :)

AdrieanKhisbe:
@MarkWesleyAtAnki, voilà! :)

AdrieanKhisbe:
@MarkWesleyAtAnki, voilà! :)

MarkWesleyAtAnki:
@AdrieanKhisbe Great - thanks!

PR #171, opened 2017-12-16
Update annotate.py

AndreiCostinescu:
Changed documentation of property "priority" of the cozmo.annotate.Annotator class such that in the official documentation the whole text will be displayed.

### DIFF
--- a/src/cozmo/annotate.py
+++ b/src/cozmo/annotate.py
@@ -227,7 +227,7 @@ class Annotator:
 
     Subclasses of Annotator handle applying a single annotation to an image.
     '''
-    #: int: The priority of the annotator - Annotators with lower numbered 
+    #: int: The priority of the annotator - Annotators with higher numbered 
     #: priorities are applied first.
     priority = 100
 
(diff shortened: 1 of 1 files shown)


AndreiCostinescu:
Hi, I have a question regarding the documentation of the priority. 
In the `cozmo.annotate.ImageAnnotator` class it is said that a higher priority will be "rendered" first (with the possibility of being overwritten by annotations with lower priorities). However, in the documentation of the `priority` property in the `cozmo.annotate.Annotator` class, it is said that 

> Annotators with lower numbered priorities are applied first.

Does the documentation have the same meaning? 
I.e. will an annotator with a low priority (small number) be applied first in the image?

If that is the case, I would suggest changing the formulation of the documentation to make it a bit clearer. Something like what I wrote above like: "Annotators with a low priority (small priority value) will be applied first on the image."

MarkWesleyAtAnki:
@AndreiCostinescu Questions about the SDK are best asked on https://forums.anki.com/ - in this case the annotators are sorted in reverse - so largest priorities first:
```python
    def _sort_annotators(self):
        self._sorted_annotators = sorted(self._annotators.values(),
                key=lambda an: an.priority, reverse=True)
```

and later applied in that sorted order, so the description at the top of the class is correct, but the later "Annotators with lower numbered priorities are applied first." is incorrect and misleading

PR #172, opened 2017-12-16
Update opengl.py

AndreiCostinescu:
Removed unnecessary quotes in the installation instruction of cozmo[3dviewer].

### DIFF
--- a/src/cozmo/opengl.py
+++ b/src/cozmo/opengl.py
@@ -25,6 +25,7 @@ Warning:
     with an implementation of GLUT (OpenGL Utility Toolkit).
 
     To install the Python packages do ``pip3 install --user cozmo[3dviewer]``
+    or ``pip3 install --user 'cozmo[3dviewer]'`` depending on your shell.
 
     On Windows and Linux you must also install freeglut (macOS / OSX has one
     preinstalled).
--- a/src/cozmo/opengl.py
+++ b/src/cozmo/opengl.py
@@ -24,7 +24,7 @@ Warning:
     This package requires Python to have the PyOpenGL package installed, along
     with an implementation of GLUT (OpenGL Utility Toolkit).
 
-    To install the Python packages do ``pip3 install --user 'cozmo[3dviewer]'``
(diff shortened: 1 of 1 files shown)


brakhane:
The quotes are necessary in some shells, like zsh.

AndreiCostinescu:
Oh, I didn't realize... In my case it didn't work with the quotes and that's why I removed them.

MarkWesleyAtAnki:
@AndreiCostinescu What OS / shell were you using that the quotes caused an issue? Note that we also use quotes around other places that pip install with square braces - e.g. cozmo[camera]:
http://cozmosdk.anki.com/docs/install-macos.html#sdk-installation
(The quotes were added when zsh users reported issues otherwise, but bash etc. are fine with and without them)

AndreiCostinescu:
Hi @MarkWesleyAtAnki 
I am using Windows 10 with the standard command prompt (cmd from the Win + r) and I get the error whenever I use quotes around the package I want installed. The square brackets are fine, but not the quotes..

src/cozmo/opengl.py
```diff
@@ -24,7 +24,8 @@
     This package requires Python to have the PyOpenGL package installed, along
     with an implementation of GLUT (OpenGL Utility Toolkit).
 
-    To install the Python packages do ``pip3 install --user 'cozmo[3dviewer]'``
+    To install the Python packages do ``pip3 install --user cozmo[3dviewer]``
```
MarkWesleyAtAnki: I tried Windows, Bash and Zsh, and double quoutes seem to work in all of them, so I'd prefer it was just changed to:
pip3 install --user "cozmo[3dviewer]"

AndreiCostinescu:
Hi @MarkWesleyAtAnki 
Is it ok if I also look for other places where the single or no quotes are used are replace them (everywhere) with double quotes?

MarkWesleyAtAnki:
@AndreiCostinescu Sure, but I think the only other case will be for the initial install of cozmo[camera] - and that's split into separate Mac, Windows and Linux pages so I think they're fine as they are (only the Mac and Linux ones have single quotes in).

PR #173, opened 2017-12-16
Update pets.py

AndreiCostinescu:
Clarity changes in the documentation of some functions.

### DIFF
--- a/src/cozmo/pets.py
+++ b/src/cozmo/pets.py
@@ -23,7 +23,7 @@ Each pet is assigned a :class:`Pet` object, which generates a number of
 observable events whenever the pet is observed, etc.
 
 If a pet goes off-screen, it will be assigned a new object_id (and
-therefore Pet object) when it returns. This is because the system can
+therefore a new Pet object will be created) when it returns. This is because the system can
 only tell if something appears to be a cat or a dog; it cannot recognize
 a specific pet or, for instance, tell the difference between two dogs.
 
@@ -50,7 +50,7 @@ from . import util
 from ._clad import _clad_to_game_anki
 
 
-#: Length of time to go without receiving an observed event before
+#: Number of seconds to go without receiving an observed event before
 #: assuming that Cozmo can no longer see a pet.
 PET_VISIBILITY_TIMEOUT = objects.OBJECT_VISIBILITY_TIMEOUT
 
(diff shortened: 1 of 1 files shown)


src/cozmo/pets.py
```diff
@@ -23,7 +23,7 @@
 observable events whenever the pet is observed, etc.
 
 If a pet goes off-screen, it will be assigned a new object_id (and
-therefore Pet object) when it returns. This is because the system can
+therefore a new Pet object will be created) when it returns. This is because the system can
```
MarkWesleyAtAnki: This makes that one line noticeably longer than the others here - please adjust this (and the subsequent 2 lines) so that they're all about 80 characters wide - e.g. add a line break after "because".
Note: slightly over 80 is fine - e.g. don't add a whole new line if it's 82 characters.

src/cozmo/pets.py
```diff
@@ -50,7 +50,7 @@
 from ._clad import _clad_to_game_anki
 
 
-#: Length of time to go without receiving an observed event before
+#: Number of seconds to go without receiving an observed event before
```
MarkWesleyAtAnki: Ideally the wording here would be similar to that in objects.py for OBJECT_VISIBILITY_TIMEOUT - adding that this is in seconds is good (that is already done in the other places), but I'm not sure there's big win of changing from:
"Length of time in seconds"
to
"Number of seconds"

AndreiCostinescu:
Hi @MarkWesleyAtAnki 
there are also other places in other classes, where the documentation of the "Length of time in seconds" thing is used. 
In some cases, the documentation also features the "(0.2)".
```
If it disappears for more than
(old) PET_VISIBILITY_TIMEOUT (0.2) seconds and then is seen again, a	
(new) PET_VISIBILITY_TIMEOUT seconds and then is seen again, a
EvtPetDisappeared will be dispatched, followed by another EvtPetAppeared event.
```

Can I change this also in the other classes or is the (0.2) really relevant? :)

MarkWesleyAtAnki:
@AndreiCostinescu Yes, please change the other 0.2 entries too - they are misleading as the time was increased to 0.4 a few months back (without remembering to update these), and it's safer to just remove it from the comment to avoid it going out of sync again.

AndreiCostinescu:
@MarkWesleyAtAnki Sure, thanks for the reply 👍

PR #199, opened 2018-07-24
COZMO-10820 Add docs and exception re: go_to_object doesn't support CustomObject

msintov:


src/cozmo/robot.py
```diff
@@ -2015,6 +2015,7 @@ def go_to_object(self, target_object, distance_from_object,
 
         Args:
             target_object (:class:`cozmo.objects.ObservableObject`): The destination object.
+                CustomObject instances are not supported.
```
shawnblakesley: Do we want to add something like "At this time"?
How much time do we think we will have to come back to Cozmo once Vector is launched?

PR #219, opened 2022-08-22
fix for issue #215, #216, and #218

SRamezanH:
a quick fix for issues #215, #216, and #218. caused by updates in asyncio.

PR #230, opened 2025-01-24
Update SDK to work with Python 3.12+

Tch1b0:
Updated two `asyncio` function calls to work with Python 3.12 and above.


enochlev (approved):

decbr1 (approved):
this is the same as PR #219, no?

Tch1b0:
Yes you're right, just noticed that too.

PR #219 fixes the same thing.

